PR #2, opened 2020-10-01
Integrate linters

NicholasNagy:
Integrating linters into the CI, and adding functionality to comment the linting report for both the Angular front-end and Python back-end.

This will help the reviewer point out areas of the code that need improvement for consistency.

### DIFF
--- a/.github/workflows/client_workflow.yml
+++ b/.github/workflows/client_workflow.yml
@@ -1,6 +1,9 @@
 name: build_client
 
-on: [push, pull_request]
+on: 
+  pull_request:
+    branches:
+      - master
 
 jobs:
   test:
--- a/.github/workflows/nglint.yml
+++ b/.github/workflows/nglint.yml
@@ -1,0 +1,45 @@
+name: NgLint
+
+on: 
+  pull_request:
+    branches:
+      - master
+
+
+jobs:
+  nglint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    steps:
+    - name: Git checkout
+      uses: actions/checkout@v2
+          
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Npm install and test
+      working-directory: client
+      run: |         
+        npm i
+        npm i -g @angular/cli
+        REPORT=$(ng lint --force=true)
+        echo "$REPORT"
+        echo ::set-env name=EXPOSED_REPORT::"${REPORT//$'\n'/'%0A%0A'}"
+    - name: Comment
+      uses: actions/github-script@v3
+      env: 
+        URL: ${{ github.event.pull_request.comments_url }}
+      with:
+        github-token: ${{secrets.GITHUB_TOKEN}}
+        script: |
+          github.issues.createComment({
+            issue_number: context.issue.number,
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            body: process.env.EXPOSED_REPORT.replace("%0A","\n")
+          })
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -21,8 +21,8 @@ jobs:
         pip install pylint
 
     - name: Run Lint
+      working-directory: server
       run: |
-        working-directory: server
         REPORT=$(pylint --exit-zero `ls -R|grep .py$|xargs`)
         echo "$REPORT"
         echo ::set-env name=EXPOSED_REPORT::"${REPORT//$'\n'/'%0A%0A'}"
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -9,7 +9,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
     - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
@@ -20,12 +19,27 @@ jobs:
         python -m pip install --upgrade pip
         pip install pylint
 
-    - name: Run Lint
-      working-directory: server
+    - name: Checkout master
+      uses: actions/checkout@v2
+      with:
+        ref: 'master'
+
+    - name: Run Lint Master
       run: |
         REPORT=$(pylint --exit-zero `ls -R|grep .py$|xargs`)
         echo "$REPORT"
-        echo ::set-env name=EXPOSED_REPORT::"${REPORT//$'\n'/'%0A%0A'}"
+        echo ::set-env name=MASTER_REPORT::"${REPORT//$'\n'/'%0A%0A'}"
+
+    - name: Checkout Branch
+      uses: actions/checkout@master
+      with:
+        ref: ${{ github.ref }}
+
+    - name: Get Diff
+      run: |
+        DIFF_REPORT=$(python3 diff_pylint.py "$MASTER_REPORT")
+        echo $DIFF_REPORT
+        echo ::set-env name=DIFF::"${DIFF_REPORT//$'\n'/'%0A%0A'}"
 
     - name: Comment
       uses: actions/github-script@v3
@@ -38,5 +52,5 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: process.env.EXPOSED_REPORT.replace("%0A","\n")
+              body: "Pylint Report \n\n" + process.env.DIFF.replace("%0A","\n")
             })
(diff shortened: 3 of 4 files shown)


.github/workflows/client_workflow.yml
```diff
@@ -1,6 +1,9 @@
 name: build_client
 
-on: [push, pull_request]
+on: 
+  pull_request:
```
RobertBeaudenon, line 4: What is the reason behind removing the GitHub action on push? Wouldn't be better to have it on push too in order to fix whatever problem we might encounter?

.github/workflows/nglint.yml
```diff
@@ -0,0 +1,45 @@
+name: NgLint
+
+on: 
+  pull_request:
```
RobertBeaudenon, line 4: What is the reason behind not adding the GitHub action on push? Wouldn't be better to have it on push too in order to fix whatever the lint is proposing?

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,42 @@
+name: Pylint
+
+on: 
+  pull_request:
```
RobertBeaudenon, line 4: What is the reason behind not adding the GitHub action on push? Wouldn't be better to have it on push too in order to fix whatever the lint is proposing?

.github/workflows/server_workflow.yml
```diff
@@ -1,6 +1,9 @@
 name: build_server
 
-on: [push, pull_request]
+on:
```
RobertBeaudenon, line 3: What is the reason behind removing the GitHub action on push? Wouldn't be better to have it on push too in order to fix whatever problem we might encounter?

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,42 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
```
RobertBeaudenon, line 21: Should we install the dependencies mentioned in the requirement.txt?

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - name: Run Lint
+      working-directory: server
+      run: |
+        REPORT=$(pylint --exit-zero `ls -R|grep .py$|xargs`)
```
RobertBeaudenon, line 26: Maybe would be nice to add a comment on top of that line on what is happening exactly for future references or for people who are interested. I know that '--exit-zero' is to treat all error as a warning but what is happening after that.

.github/workflows/nglint.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Npm install and test
+      working-directory: client
+      run: |         
+        npm i
+        npm i -g @angular/cli
```
RobertBeaudenon, line 29: Why don't you install the dependencies before as you did in the pylint workflow to be consistent.

.github/workflows/nglint.yml
```diff
@@ -0,0 +1,45 @@
...
+        REPORT=$(ng lint --force=true)
+        echo "$REPORT"
+        echo ::set-env name=EXPOSED_REPORT::"${REPORT//$'\n'/'%0A%0A'}"
+    - name: Comment
```
RobertBeaudenon, line 33: What's that part doing, is it adding the report to Github? and if so is it under the pull request?

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,42 @@
...
+        echo "$REPORT"
+        echo ::set-env name=EXPOSED_REPORT::"${REPORT//$'\n'/'%0A%0A'}"
+
+    - name: Comment
```
RobertBeaudenon, line 30: What's that part doing, is it adding the report to Github? and if so is it under the pull request?

RobertBeaudenon (commented):
I added some comments to get some clarification on certain decisions but other than that looks good.

.github/workflows/client_workflow.yml
makchamp: yes i agree

makchamp (commented):
Thank you for adding the linters. I have 2 suggestions:
- Amend it into the existing workflows file into the existing workflows files
- Maybe show a quick diff instead of the error log (since that can be accessed directly from the build)
this would probably reduce the size of the comments as well. 
This is a good alternative if we don't want to fail the build on any linting errors. We also have the option to limit exactly what errors we want to be triggered later on.

NicholasNagy:
> * Amend it into the existing workflows files so we dont have more builds running unnecessarily

Do you want it to run before the tests, so that if the tests fail, the linting reports will still be posted? Or after so that they don't get posted if the tests fail?

.github/workflows/client_workflow.yml
NicholasNagy: So right now, with the new configuration, it will only run on pushes to a Pull Request, and when the Pull Request is opened. Previously, it would run on that and on every push to github as well. What do you guys prefer?

.github/workflows/nglint.yml
NicholasNagy: So as mentioned above, it will run on a push, but only if pushing to an open pull request

.github/workflows/pylint.yml
NicholasNagy: So in this configuration, we actually can't change it. You need to have the lint run only on a pull request, because or else the build will fail as it won't have a PR to push to

.github/workflows/pylint.yml
NicholasNagy: It's not necessary in order to lint. We just need the linter and the source files.

.github/workflows/nglint.yml
NicholasNagy: I install npm packages here, because it's actually needed to run the angular linter.

.github/workflows/nglint.yml
NicholasNagy: Line 30 saves the output of the lint to the REPORT variable.
line 31 prints it out so that we can view the output in the CI if we want.
line 32 sets the environment variable EXPOSED_REPORT to the value of REPORT, but with the newline character replaced with %0A%0A. The reason for the switch is that when setting an environment variables (or more broadly when using the github actions toolkit) it will escape the newline character out, effectively deleting them when commenting on the pull request.

.github/workflows/pylint.yml
NicholasNagy: I tried to answer this question above (the same should apply here)

makchamp:
> 
> 
> > * Amend it into the existing workflows files so we dont have more builds running unnecessarily
> 
> Do you want it to run before the tests, so that if the tests fail, the linting reports will still be posted? Or after so that they don't get posted if the tests fail?
> 
> **EDIT** : Actually, I just realized that if we want to combine the linters with the build workflows, then I cannot change those workflows to build on any push (as we discussed), as a push to the server would fail if a pull request is not opened for that branch. The reason being is that the github action for commenting would fail as it would not have a reference to the pull request to comment on.

Try this: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif

```
steps:
 - name: pr comment
   if: ${{ github.event_name == 'pull_request'}}
   run: echo comment
```

makchamp:
> 
> 
> > * Amend it into the existing workflows files so we dont have more builds running unnecessarily
> 
> Do you want it to run before the tests, so that if the tests fail, the linting reports will still be posted? Or after so that they don't get posted if the tests fail?
> 
I'd run this after the tests as they have less priority. You can chose to run them regardless if the actual tests pass or fail.

NicholasNagy:
> > > * Amend it into the existing workflows files so we dont have more builds running unnecessarily
> > 
> > 
> > Do you want it to run before the tests, so that if the tests fail, the linting reports will still be posted? Or after so that they don't get posted if the tests fail?
> > **EDIT** : Actually, I just realized that if we want to combine the linters with the build workflows, then I cannot change those workflows to build on any push (as we discussed), as a push to the server would fail if a pull request is not opened for that branch. The reason being is that the github action for commenting would fail as it would not have a reference to the pull request to comment on.
> 
> Try this: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif
> 
> ```
> steps:
>  - name: pr comment
>    if: ${{ github.event_name == 'pull_request'}}
>    run: echo comment
> ```

Oh yeah! I forgot about using if statements in the workflow, that's a good point, I will try it out tonight if I have time.

NicholasNagy:
Accidentally merged this commit though the command line and I have legitimately no idea how I did it.

PR #3, opened 2020-10-01
Issue #12 System admin account

RobertBeaudenon:
Provided the following features for the System Admin:
- Registration API
- Login API
- Logout API
- Token authentication
- Password Hashing using SHA256


server/django_server/settings.py
```diff
@@ -34,23 +34,38 @@
...
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
 
+    # Local apps
+    'user_account',
+
+    # 3rd party
     'rest_framework',
+    'rest_framework.authtoken',
```
NicholasNagy, line 62: What do you mean by 3rd party?

server/django_server/settings.py
```diff
@@ -95,23 +109,34 @@
...
+PASSWORD_HASHERS = (
+    'django.contrib.auth.hashers.PBKDF2PasswordHasher',
+    'django.contrib.auth.hashers.PBKDF2SHA1PasswordHasher',
+    'django.contrib.auth.hashers.BCryptSHA256PasswordHasher',
+    'django.contrib.auth.hashers.BCryptPasswordHasher',
+    'django.contrib.auth.hashers.SHA1PasswordHasher',
+    'django.contrib.auth.hashers.MD5PasswordHasher',
+    'django.contrib.auth.hashers.CryptPasswordHasher',
+)
+
```
NicholasNagy, line 122: Do we need all of these?

Sorry if this is a noob question, but I have the impression that we would only use like 1 or 2 hashers

server/django_server/settings.py
```diff
@@ -95,23 +109,34 @@
...
+#     {
+#         'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
+#     },
+#     {
+#         'NAME': 'django.contrib.auth.password_validation.CommonPasswordValidator',
+#     },
+#     {
+#         'NAME': 'django.contrib.auth.password_validation.NumericPasswordValidator',
+#     },
+# ]
```
NicholasNagy, line 139: Why are we commenting this out?

server/django_server/settings.py
RobertBeaudenon: It's just to separate what app we develop internally and what app do we use from outside to organize the INSTALLED APP array in the settings.

server/user_account/models.py
```diff
@@ -0,0 +1,34 @@
...
+
+class CustomUser(AbstractBaseUser):
+    USERS = (
+        ('SA', 'System Admin'),
+        ('IM', 'Inventory Manager'),
+        ('SK', 'Stock Keeper'),
+    )
+    user_name = models.CharField(max_length=30, unique=True)
+    first_name = models.CharField(max_length=30)
+    last_name = models.CharField(max_length=30)
```
NicholasNagy, line 17: Why a max length of 30?

server/user_account/models.py
```diff
@@ -0,0 +1,34 @@
...
+    role = models.CharField(max_length=2, choices=USERS)
```
NicholasNagy, line 18: Why max length of 2?

server/user_account/models.py
```diff
@@ -0,0 +1,34 @@
...
+    is_active = models.BooleanField(default=True)
+    email = models.EmailField(verbose_name='email', max_length=255, unique=True)
```
NicholasNagy, line 20: Why is this a max length of 255, but the names are 30?

server/user_account/serializers.py
```diff
@@ -0,0 +1,31 @@
+from rest_framework import serializers
+from .models import CustomUser
+
+
+class UserSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CustomUser
+        fields = ['id', 'user_name', 'first_name', 'last_name', 'password', 'role', 'email']
+
+    # overriding save method
```
NicholasNagy, line 10: Great, can you explain why you are overriding it though?

server/user_account/tests.py
```diff
@@ -0,0 +1,17 @@
...
+from .models import CustomUser
+
+
+class CustomUserTestCase(TestCase):
+    def setUp(self):
+        CustomUser.objects.create(user_name="test_user", email="test@test.com", id=1, first_name='test',
+                                  last_name='user', role='SA')
+
+    def test_user_creation(self):
+        """ User was created correctly """
```
NicholasNagy, line 11: Can you include this before the method the definition?

server/user_account/views.py
```diff
@@ -0,0 +1,71 @@
...
+from rest_framework import generics
+from django.contrib import auth
+from django.contrib.auth.hashers import check_password
+
+
+class RegistrationView(viewsets.ModelViewSet):
+    queryset = CustomUser.objects.all()
+    serializer_class = UserSerializer
+
+    def create(self, request, *args, **kwargs):
```
NicholasNagy, line 17: Can you write a small comment explaining this method?

server/user_account/views.py
```diff
@@ -0,0 +1,71 @@
...
+                    token = Token.objects.create(user=user)
+
+                data = {'user': username, 'token': token.key}
+                return Response(data, status=status.HTTP_200_OK)
+            else:
+                return Response({'detail': 'Invalid credentials'}, status=status.HTTP_401_UNAUTHORIZED)
+                pass
+        except CustomUser.DoesNotExist:
+            return Response({'detail': 'Invalid user'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+            pass
```
NicholasNagy, line 56: what's the purpose of pass here?

server/user_account/views.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class LogoutView(generics.GenericAPIView):
+    def post(self, request):
+        return self.remove_token(request)
+
+    def remove_token(self, request):
```
NicholasNagy, line 63: Can you comment on what this method does?

NicholasNagy (changes requested):
I've left a few comments, explaining my thoughts, but there seems to be 2 main things missing in my view point.

- You are missing comments explaining what your classes are for
- You are missing tests for logging in, logging out and registration.

I'd like to meet with you soon to discuss how the authentication works.

server/django_server/settings.py
RobertBeaudenon: Yes i just forgot to remove them.

server/django_server/settings.py
RobertBeaudenon: Because we didn't yet define what restriction do we want on passwords, so for testing purposes, I just commented them in order to create whatever password I want.

server/user_account/models.py
RobertBeaudenon: We can redefine them, but I didn't think that we would have names bigger than 30 characters, I can redefine it to what we decide.

server/user_account/models.py
RobertBeaudenon: Because I specify that each different user role is either SA, IM, SK respectively for system admin, inventory manager, and stock keeper.

server/user_account/models.py
RobertBeaudenon: Because emails are usually longer and its a combination of first name + last name + whatever, but again we can agree on another number.

server/user_account/serializers.py
RobertBeaudenon: I'm overriding it to add hashing to the password through the set_password() method. If I don't do that the elements will be saved as they are. So once I override this method I can manipulate whatever was sent in the request before storing them in the database.

server/user_account/tests.py
RobertBeaudenon: The convention is to have it inside the method as I did.

RobertBeaudenon: Should I proceed with the convention or write them on top

server/user_account/views.py
RobertBeaudenon: Yeah I should probably remove it, It does not make sense to have it.

server/user_account/views.py
RobertBeaudenon: Yes, I will.

server/user_account/views.py
RobertBeaudenon: So when a user logout I delete the token related to that user from the database.

server/django_server/settings.py
NicholasNagy: okay makes sense, but why don't we put the rest framework portion with the rest of django?

server/django_server/settings.py
NicholasNagy: Can you add a todo to explain this so that we don't forget?

server/user_account/models.py
NicholasNagy: Makes sense.

server/user_account/models.py
NicholasNagy: okay, maybe we can discuss with Ajay what he might expect, or we can do our own research to see the norm.

server/user_account/serializers.py
NicholasNagy: I mean in the comment for the method

server/django_server/settings.py
RobertBeaudenon: In order to not get mixed up with what Django comes with and what we installed in order to remove them if we decide to not use them anymore

server/user_account/tests.py
NicholasNagy: Link me the convention?

server/user_account/views.py
NicholasNagy: I meant in the comment above

server/user_account/serializers.py
RobertBeaudenon: Oh I see what you mean, yes I can add that.

server/user_account/tests.py
RobertBeaudenon: It's the PEP 257 Docstring Conventions -> https://www.python.org/dev/peps/pep-0257/

NicholasNagy: [https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
I guess I assumed Python had Java-like method commenting syntax, my bad.

server/user_account/models.py
NicholasNagy: [https://stackoverflow.com/questions/20958/list-of-standard-lengths-for-database-fields](https://stackoverflow.com/questions/20958/list-of-standard-lengths-for-database-fields)

Let's just use a safe 255 characters. I think that would be our safest choice

server/user_account/views.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        Deleting user token from the database when he logout.
+        :param request
+        """
+
+        try:
+            Token.objects.get(user=request.user).delete()
+        except Token.DoesNotExist:
+            return Response({'detail': 'Invalid Token'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+            pass
```
NicholasNagy, line 100: Did you forget to remove this pass as well?

NicholasNagy (approved):
Everything looks good, now.

I'll merge!

PR #6, opened 2020-10-04
Integrate linters

NicholasNagy:
Integrating linters into the CI, and adding functionality to comment the linting report for both the Angular front-end and Python back-end.

This will help the reviewer point out areas of the code that need improvement for consistency.

**EDIT:** @makchamp Right now, build times are approximately 30s for the linters. After realizing this, I think the best solution would be to have them run in parallel to not slow down builds. Build times might not be so bad now, but they won't scale unless we parallelize every chance we can. Let me know your thoughts, so that I can work either finish up the angular linting workflow or integrate them into the current builds.


makchamp (approved):
> EDIT: @makchamp Right now, build times are approximately 30s for the linters. After realizing this, I think the best solution would be to have them run in parallel to not slow down builds. Build times might not be so bad now, but they won't scale unless we parallelize every chance we can. Let me know your thoughts, so that I can work either finish up the angular linting workflow or integrate them into the current builds.


Nice job on this I like the diff.
As to your point on the build times, the actual linting is usually a quick process. You can test it on other big sized projects to verify. The majority of that 30 seconds is going to dedicated to setting up the environment. So if we combine it, it only has to be done once. That being said, it is no problem the way it is right now given the size of the yaml linting files. It is clearly better for clarity to have them separate. However, keep in mind that there will be no linting executed on normal pushes with this setup.

diff_pylint.py
```diff
@@ -0,0 +1,11 @@
+import subprocess
```
makchamp, line 1: I would move this file into the .workflows directory

NicholasNagy: done.

NicholasNagy:
@makchamp So I reverted the change that removes the build on push (now it will build on push and on pull request like before)

I kept the linting separate, simply because we actually have to build the project twice to get the diffs for Angular, and since these workflows a large and complex, it's best to keep them separate for readability.

If all is well, this PR should be ready to merge!

PR #7, opened 2020-10-04
System admin dashboard

Averynder:
This will add the small dashboard / listing of options for a system admin into production for the Angular front end. The options will be displayed in an unordered list and currently when a user clicks on one of the list items, the header above the options will change to the option clicked.


NicholasNagy (approved):
Looks good, I have no feedback for this change.

This is related to epic 1 #31 .

client/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(component).toBeTruthy();
+  });
+
+  // Default Option
+  it('#onSelect should set #selectedOption to "The First option"', () => {
+    expect(component.selectedOption.title).toBe('Create My Account', 'Initial Option');
+  });
+
+  // Tests Options are clickable
+  it('should test click 2nd option', () => {
```
makchamp, line 32: I think this test name is wrong. I see you are clicking on all elements and not just the second one.

makchamp (changes requested):
Good and simple change. Just a few things:
- You must declare the dashboard component in the parent as well (which is AppComponent). Not doing this will cause the tests to complain about 'not know element' 
So in app.component.spec.ts, you can add the declaration there
- We can now delete the demo component and demo service. This will prevent other branches from merging into it again (since i think @RobertBeaudenon removed/replaced it from his)

client/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    const elements = fixture.debugElement.queryAll(By.css('li'));
+    elements.forEach(element => {
+      fixture.detectChanges();
+      element.nativeElement.click();
+      expect(component.onSelect).toHaveBeenCalled();
+      theSpy.calls.reset();
+    });
+  });
+
+  // Tests 2nd option was clicked
```
makchamp, line 43: change comment here too

client/src/app/dashboard/dashboard.component.spec.ts
Averynder: You are correct, the code was changed to test them all, comment has been removed in latest commit as per your request:
c2cb3c8c06ac7c0e18bbbba34240c8047d1f068e

client/src/app/dashboard/dashboard.component.spec.ts
Averynder: comment has been removed, title has been updated as per commit: c2cb3c8c06ac7c0e18bbbba34240c8047d1f068e

makchamp (approved):
Changes look good. Thanks @Averynder 
I will refrain from merging now in case there is an issue with the current master branch.
@NicholasNagy  and @RobertBeaudenon  can decide on that tomorrow.

PR #36, opened 2020-10-10
Fix pylint workflow + clean up linter errors for django app

makchamp:
#### I modified the pylint workflow to be more accurate and produce relevant output: 
The previous linting workflow would only show module not found errors.
These are clearly false positives. It was also not detecting actual errors. 

It was discovered that pylint has a specific django lint plugin called pylint-django
The linter now runs on files output from (git ls-files '*.py') meaning tracked python files only 
The working directory has been changed to server
Also for the linter to be more accurate all packages must be installed (since there are other dependencies required)

#### I added the pylint config file .pylinrc
- Inside I add rules to ignore:
 migrations files and the manage.py script
- ignore missing docstring errors
- changed the max inheritance depth to 15 (I think the limit is only 7 which is impossible to keep when using some classes like Views)
#### I cleaned up all the previous linting errors from the django app
- I did not fix all errors. Some I simply added a comment to ignore where I thought most appropriate. 

**This is how I would run the linter locally:**
From inside the server directory:
pylint --load-plugins pylint_django $(git ls-files '*.py')

Note: I added autopep8 as a dependency. It can be removed if most team members are using something else (like whatever pycharm uses for formatting)


NicholasNagy:
Can you comment on the lines you ignored? It's kind of hard to understand what is being ignored and whether it is critical or not, based on how the line is being ignored.

server/user_account/models.py
```diff
@@ -32,6 +33,7 @@ def get_role(self):
 
     @receiver(post_save, sender=settings.AUTH_USER_MODEL)
     def create_auth_token(sender, instance=None, created=False, **kwargs):
+        # pylint: disable-all
```
makchamp, line 36: I disable all here because the method does not seem complete and will probably be modified later. 
Here are the errors:
user_account/models.py:35:4: E0213: Method should have "self" as first argument (no-self-argument)
user_account/models.py:35:0: W0613: Unused argument 'kwargs' (unused-argument)
user_account/models.py:35:4: R0201: Method could be a function (no-self-use)

server/user_account/serializers.py
```diff
@@ -5,11 +5,13 @@
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = CustomUser
-        fields = ['id', 'user_name', 'first_name', 'last_name', 'password', 'role', 'email', 'is_active']
+        fields = ['id', 'user_name', 'first_name', 'last_name',
+                  'password', 'role', 'email', 'is_active']
 
-    def save(self):
+    def save(self): # pylint: disable=arguments-differ
```
makchamp, line 11: Arguments differ from the overridden method. Ignored because it's just 'self'

server/user_account/tests.py
```diff
@@ -90,11 +125,14 @@ def setUp(self):
 
     def api_authentication_valid_token(self):
         """ Added valid token header to request """
+        # pylint: disable=no-member
```
makchamp, line 128: The member 'credentials' is not being detected. I tested it and the code is indeed working and this call is required for the test to work.

makchamp (commented):
> Can you comment on the lines you ignored? It's kind of hard to understand what is being ignored and whether it is critical or not, based on how the line is being ignored.

I added them as review comments

server/user_account/models.py
NicholasNagy: Then can you add a todo so that we don't forget?

server/user_account/serializers.py
NicholasNagy: I don't get it, so in the method that we're overriding, what are the arguments supposed to be?

server/user_account/tests.py
NicholasNagy: So are we supposed to be passing credentials somehow? I don't get why pylint is flagging this instance?

server/user_account/serializers.py
makchamp: No arguments

server/user_account/tests.py
makchamp: Its just a method for the inherited class. Probably a bug or something with pylint.

server/user_account/serializers.py
NicholasNagy: I looked it up, and the original definition is the following:

`def save(self, **kwargs):`

server/user_account/tests.py
NicholasNagy: [Seems to be a false positive](http://pylint-messages.wikidot.com/messages:e1101)

NicholasNagy (commented):
In addition to my comment about the save method, please rebase and rename your commits to follow our current commit convention. Afterwards, I will approve and merge.

NicholasNagy (approved):
Looks good now, thanks for the PR!

I will merge!

PR #37, opened 2020-10-11
New app navigation, components and updated routing

makchamp:
#### This PR introduces new side menu navigation to the app
A side drawer has been added. It can be toggled using a menu button at the top left.
The dashboard, create members and modify members fields are currently clickable.
Everything after the login page is now inside the HomeComponent. The app structure is like so:
HOME COMPONENT => CHILD COMPONENTS { DashboardComponents, ManageTeamsComponent, AuditComponent, etc.. } 

**The default (empty)  route is now the home component.**
As the default route was used for the signin/login page, a new 'login' route was created as a replacement
Child routes of the home component are all selectable side menu options.
The empty route redirects to the dashboard component by default.
This route still cannot be accessed unless authenticated.

**Added a toolbar and made the layout responsive**
I followed the guidlines for the layout from the angular docs:
https://material.angular.io/components/sidenav/overview


client/src/app/app.module.ts
```diff
@@ -17,7 +16,6 @@ import { CookieService } from 'ngx-cookie-service';
     BrowserModule,
     BrowserAnimationsModule,
     AuthModule,
-    AuthRoutingModule,
```
RobertBeaudenon, line 20: Why did you delete the routing of auth? keep it even if there is only one route inside we might add other after related to authentication. And we don't want to mix contexts.

client/src/app/components/home/home.component.html
```diff
@@ -1,2 +1,18 @@
-<app-dashboard></app-dashboard>
-<button (click)="logout()">Logout</button>
```
RobertBeaudenon, line 2: Why did you delete the logout button? Keep it for test purposes until we add it inside the toolbar. Or just create the toolbar using angular material and add it in it.

client/src/app/modules/alta-main-routing/alta-main-routing.module.ts
```diff
@@ -1,18 +1,31 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
+import { AuthTabsComponent } from 'src/app/components/auth-tabs/auth-tabs.component';
+import { DashboardComponent } from 'src/app/components/dashboard/dashboard.component';
 import { HomeComponent } from 'src/app/components/home/home.component';
+import { ManageMembersComponent } from 'src/app/components/manage-members/manage-members.component';
 import { AuthGuard } from '../../guards/auth.guard';
 
 const routes: Routes = [
+  { path: 'login', component: AuthTabsComponent },
```
RobertBeaudenon, line 11: As mentioned above, I think it would be better to recreate the file responsible for routing for authentication stuff. So when you do that please add this route in the other file.

client/src/app/modules/alta-main/alta-main.module.ts
```diff
@@ -1,15 +1,26 @@
...
+  declarations: [
+    HomeComponent,
+    DashboardComponent,
+    SideNavComponent],
+  imports: [
+    CommonModule,
+    MatModule,
+    ReactiveFormsModule,
+    HttpClientModule,
+    RouterModule
```
RobertBeaudenon, line 26: You forgot to add inside your imports the angular material-specific modules that you are using. I would say do not import the whole module library, just do it by a specific module(like I did in the auth module). Just check if its better for performance to just add it by subcomponents that you use.
Screen Shot 2020-10-11 at 7 09 14 PM

client/src/app/components/home/home.component.html
```diff
@@ -1,2 +1,18 @@
...
+<mat-drawer-container class="home-container">
+
+    <mat-drawer #drawer opened="true" mode="side">
+        <app-sidenav></app-sidenav>
+    </mat-drawer>
+
+    <mat-drawer-content class="home-content">
+        <button type="button" mat-button (click)="drawer.toggle()">
```
RobertBeaudenon, line 8: The button is still not a mat-button by looking at its style in the page.
Screen Shot 2020-10-11 at 7 15 33 PM

client/src/app/services/token-interceptor.ts
```diff
@@ -37,9 +37,10 @@ export class TokenInterceptor implements HttpInterceptor {
     // If token is available in the cookie
     if (token) {
       // We set the Authorization header: token inside the object for every request
-      headersConfig['Authorization'] = `Token ${token}`;
+      const authHeader = 'Authorization';
+      headersConfig[authHeader] = `Token ${token}`;
```
RobertBeaudenon, line 41: you are doing that because the lint is complaining?

client/src/app/modules/material/material-module.ts
```diff
@@ -0,0 +1,14 @@
...
+  declarations: [],
+  imports: [
+    MatSidenavModule,
+  ],
+  providers: [],
+  exports: [
+    MatSidenavModule,
+  ]
+})
+export class MatModule { }
```
RobertBeaudenon, line 14: What is the reason behind creating this module?

client/src/app/modules/alta-main-routing/alta-main-routing.module.ts
```diff
@@ -1,18 +1,33 @@
...
     canActivate: [AuthGuard], // If user has a valid token he will be able to access comment page
+    children: [
+      {
+        path: '',  // Empty path needed in child too to be able to force a redirect
+        redirectTo: 'dashboard',
+        pathMatch: 'full'
+      },
+      { path: 'dashboard', component: DashboardComponent },
+      { path: 'create-members', component: CreateMembersComponent },
+      { path: 'modify-members', component: ModifyMembersComponent }
```
RobertBeaudenon, line 24: Are you adding a route for the rest of the list in the submenu? Maybe add a TODO on the side.

RobertBeaudenon (changes requested):
I added some comments to the code, to understand why some decisions were made.

client/src/app/app.module.ts
makchamp: I did not think it was necessary because it was only 1 route and to be able to see the whole picture at once. I can add I back if you think we will add more routes there later.

client/src/app/components/home/home.component.html
makchamp: yes i was going to add it to the toolbar

client/src/app/modules/alta-main/alta-main.module.ts
makchamp: I wanted to import all the mat modules in one place for simplicity. If the module is big it will effect build time not performance. You think the module will get big enough for it to matter ?

client/src/app/components/home/home.component.html
makchamp: I think I will just change it to a regular button since it is only temporary anyway.

client/src/app/services/token-interceptor.ts
makchamp: Yes it was complaining locally. After this fix it stopped but now it is complaining from the workflow. I dont know why.

client/src/app/modules/alta-main-routing/alta-main-routing.module.ts
makchamp: No only for modify and create member components for now

client/src/app/app.module.ts
RobertBeaudenon: Yeah, I would just prefer to keep it in a separate route folder.

client/src/app/components/home/home.component.html
RobertBeaudenon: perfect.

client/src/app/modules/alta-main/alta-main.module.ts
RobertBeaudenon: So I've looked a bit into it, I think since we are going to use angular material all across the website, we should probably create a specific module that will just be responsible of importing the particular components of angular material that we are using, and then we would be able to import this particular module inside the other modules that need it. So this way it would make the code more organized.
https://stackoverflow.com/questions/57038161/angular-8-is-importing-all-material-modules-into-the-main-app-module-better-th

client/src/app/modules/material/material-module.ts
RobertBeaudenon: So as I mentioned in the conversation about the angular material module, what we can do from now on is add all of our angular material used inside this module, and then if we need those components in other particular modules such as auth-module and alta-main module we would just have to import this MatModule which will be centralizing the import of stuff related to angular material. What do you think about that?

client/src/app/services/token-interceptor.ts
RobertBeaudenon: Perfect 👍

client/src/app/components/home/home.component.ts
```diff
@@ -1,20 +1,46 @@
...
+import { MatSidenav } from '@angular/material/sidenav';
 import { TokenService } from 'src/app/services/token.service';
 import { Router } from '@angular/router';
+import { SidenavService } from 'src/app/services/sidenav.service';
 
 @Component({
   selector: 'app-home',
   templateUrl: './home.component.html',
   styleUrls: ['./home.component.css'],
+  providers: [SidenavService],
```
RobertBeaudenon, line 12: I would say don't add the SidnavService inside the home component, do it inside the Alta-main-module.

makchamp: Yes nice catch. This is actually not needed at all. I removed it with 6589be7856d36674cec240568ad3eb202f7d177d

client/src/app/modules/material/material-module.ts
makchamp: Yes exactly. That was my reasoning.

client/src/app/modules/alta-main/alta-main.module.ts
makchamp: Yes exactly. That was my reasoning. The module is indeed MatModule

client/src/app/components/home/home.component.ts
RobertBeaudenon: But you forgot to add it to the alta-main-module inside the providers decorator.

makchamp: Yes because its not needed there as well

client/src/app/services/sidenav.service.ts
```diff
@@ -0,0 +1,29 @@
...
+import { MatDrawerToggleResult, MatSidenav } from '@angular/material/sidenav';
+
+@Injectable({
+  providedIn: 'root',
+})
+@Injectable()
+export class SidenavService {
+  private sidenav: MatSidenav;
+
+
```
RobertBeaudenon, line 12: Can you delete those empty lines.

client/src/app/services/sidenav.service.ts
```diff
@@ -0,0 +1,29 @@
...
+  public setSidenav(sidenav: MatSidenav): void {
+    this.sidenav = sidenav;
+  }
+
+  public open(): Promise<MatDrawerToggleResult> {
+    return this.sidenav.open();
+  }
+
+
```
RobertBeaudenon, line 21: Can you delete this empty line.

client/src/app/components/home/home.component.ts
```diff
@@ -1,20 +1,45 @@
...
+
+  ngOnDestroy(): void {
+    this.mobileQuery.removeEventListener('change', this.mobileQueryListener);
+  }
+
+  ngAfterViewInit(): void {
+    this.sidenavService.setSidenav(this.sidenav);
+  }
 
   logout(): void {
```
RobertBeaudenon, line 40: Can you please delete this method since now it's implemented inside toolbar component.

client/src/app/components/sidenav/sidenavListing.ts
```diff
@@ -0,0 +1,20 @@
...
+      { title: 'Create Template', routerLink: 'create-template', subMenuOptions: [] },
+    ]
+  },
+  {
+    title: 'Manage Members', routerLink: 'manage-members', subMenuOptions: [
+      { title: 'Create Members', routerLink: 'create-members', subMenuOptions: [] },
+      { title: 'Modify Members', routerLink: 'modify-members', subMenuOptions: [] },
+    ]
+  },
+
```
RobertBeaudenon, line 19: Can you please delete this empty line

client/src/app/app.module.ts
```diff
@@ -17,9 +17,9 @@ import { CookieService } from 'ngx-cookie-service';
...
     AltaMainModule,
     AltaMainRoutingModule,
+    AuthRoutingModule,
     HttpClientModule,
```
RobertBeaudenon, line 23: I think you forgot to add here the module related to angular material that you just created

RobertBeaudenon (changes requested):
I added some little comments.
You forgot to add a file to test the initialization of the side-nav service.
But overall looks good, nice job.

client/src/app/components/home/home.component.ts
makchamp: This is actually needed here as the sidenav element is inside the home component.

RobertBeaudenon: No i'm talking about the logout().

makchamp: yes sorry

RobertBeaudenon:
Thank you for addressing my comments.
Everything looks good, I will merge and we can discuss where we go from here.

PR #38, opened 2020-10-12
Obtain Clients

Averynder:
This pr merges the ability to obtain all the data from the database to the front end via JSON. It is displayed by clicking the modify client role within the current dashboard.


Averynder:
> 
> 
> You might want to add this component in the module alta-main, we are having multiple modules, the parent one is the app module then you have auth module for authentication and alta-main for when you passed authentication

This was fixed in commit 3f5576f and the PR had not yet been made

client/src/app/services/dashboard.service.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      email: 'a@gmail.com',
+      first_name: 'a',
+      last_name: 'a',
+      role: 'SA'
+    };
+
+    service.getAllClients()
+      .subscribe(jsonData => {
+        const data = JSON.parse(jsonData);
+        expect(data.user_name).toEqual('a');
```
makchamp, line 40: Maybe access the mockUsers's members instead of typing the strings out again (like you did in the other test) ?
Like this:  mockUser.user_name

client/src/app/services/dashboard.service.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+        expect(data.email).toEqual('a@gmail.com');
+        expect(data.first_name).toEqual('a');
+        expect(data.last_name).toEqual('a');
+        expect(data.role).toEqual('SA');
+      });
+
+    // This is where the request should be
+    const req = httpTestingController.expectOne('http://localhost:8000/getAllClients/');
```
makchamp, line 48: You can use BASE_URL that you created from the service file that is related to this test. Or if you dont want to make it a member extract it to a variable again. 
Its because later we might think to extract this url to a env file like in the django app

client/src/app/components/modify-client-role/modify-client-role.component.ts
```diff
@@ -0,0 +1,57 @@
...
+        for (const value of this.querrysett)
+        {
+          this.users.push(new CustomUser(value.fields));
+        }
+      });
+  }
+
+}
+
+class CustomUser
```
makchamp, line 41: We had discussed the utility of this class. I think it can be removed with some minor tweaks to the obtainClients() method. I'll look into it.

makchamp: I moved this class to the models directory. I then refactored the request to return an array of Users directly. This cut done alot of the code from inside the obtainClients() method right above this. 1303838439ed5deedd2c6edad9442ee3883d0edf
Also I made it so only relevant fields are returned from the request (no primary key or password for example) 5cceee12eb884b2d49901094c825eed86230e309

makchamp:
Thanks @Averynder for the changes.

Issue #33, opened 2020-10-06
User Story 1.4


makchamp:
### User Story: ALTA-1.4
As a system administrator, I want to be able to create my client’s employee’s user account so that I can grant their employee access to ALTA. 

### Description

- An organization must be created before creating an employee account

- Possible employee roles: inventory manager or stock keeper

- In discussion: import employee data from an Excel file for multiple account creation

### Acceptance Criteria (AT-1.4 #134)
Given that the user is logged in as a system administrator and is viewing the _Manage Organizations_ page,

1. The user is able to select an organization by clicking on its respective row in the presented table. 

2. The user selects the _Employees_ side menu option.

3. The user is presented with a table displaying the affiliated employee user accounts and a _Create_ button. 

4. The user selects the _Create_ button to display an empty form for the user to input the new user’s information.

5. Once the form is complete, the user submits the data by clicking the _Save_ button. The user is then redirected back to the table to validate the creation of the user.

### Related Mockups

- System administrator logs into ALTA
![image](https://user-images.githubusercontent.com/19229712/100304623-21351600-2f6d-11eb-9f5f-6fb469b27410.png)

- Table of subscribed organizations
![image](https://user-images.githubusercontent.com/19229712/100404495-83564f80-302f-11eb-89c7-e35d2ee2de91.png)

- Table of selected organization's users
![image](https://user-images.githubusercontent.com/19229712/100404659-ea740400-302f-11eb-8757-e3e26cc4e42a.png)

- Form to create a new client's employee user account
![image](https://user-images.githubusercontent.com/19229712/100305215-97864800-2f6e-11eb-9b1e-1104f58572d6.png)

PR #40, opened 2020-10-14
Issue #33 Authenticated employee account creation + open registration endpoint

makchamp:
**Added functionality to create employee accounts of different roles (inventory manager, stock keeper, system admin)** 
- Added the signup component (and form) within the create-members component. The form appears when you click on Create Members from the side menu. 
- The form has been extended to include a drop down menu to select the account role.
- The value from this drop down is required

**Added authenticated registration**
- The registration endpoint has been modified to check for permissions using the IsAuthenticated permission class. This means that nobody can create an account from outside (not logged in) the system (since they are not authenticated yet).
- However there are currently no limitations on creation yet once inside the system (any user can create any other user role)
- Added a TODO for the last point

**Added an open registration endpoint**
- This endpoint is purely for testing/development purposes. It allows us to create accounts from outside (not logged in)  the system. That is the only place where it is called.

### DIFF
--- a/server/user_account/tests.py
+++ b/server/user_account/tests.py
@@ -4,6 +4,7 @@ from django.test.client import Client as HttpClient
 from rest_framework import status
 from rest_framework.test import APITestCase
 from rest_framework.authtoken.models import Token
+from rest_framework.test import APIClient
 from .models import CustomUser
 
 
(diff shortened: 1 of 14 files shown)


server/user_account/tests.py
```diff
@@ -79,19 +79,19 @@ def test_registration_success(self):
...
-        response = self.client.post("/registration/", data)
+        response = self.client.post("/open-registration/", data)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
     def test_registration_unauthorized_request(self):
         """ User can't access the GET method at this particular endpoint """
-        response = self.client.get("/registration/")
+        response = self.client.get("/open-registration/")
         self.assertEqual(response.status_code,
                          status.HTTP_405_METHOD_NOT_ALLOWED)
```
RobertBeaudenon, line 97: Can you please add a test case for the new registration view you implemented. Also don't forget to test for users who don't have a token when doing the request.

server/user_account/views.py
```diff
@@ -6,14 +6,59 @@
...
+        last_name, user_name, password, role, email, is_active
+        :return: user_name, token
+        """
+        # Retrieve the authenticated user making the request
+        auth_content = {
+        'user': str(request.user),
+        'auth': str(request.auth),
+        }
+        # If not authenticated return 401
+        if auth_content['auth'] is None:
```
RobertBeaudenon, line 37: You don't need to add this check. This is handled by permission_classes=[IsAuthenticated]. So you can delete it. But just test it before in this particular case because we are overriding the method.

server/user_account/views.py
```diff
@@ -6,14 +6,59 @@
...
+            return Response({'unauthorized': 'Unauthorized request'},
+            status=status.HTTP_401_UNAUTHORIZED)
+
+        auth_user = CustomUser.objects.get(user_name=auth_content['user'])
+        # TODO: Limit account creation by role
+        if auth_user.role != 'SA':
```
RobertBeaudenon, line 43: Good catch, we will need to do that later.

server/user_account/views.py
```diff
@@ -6,14 +6,59 @@
...
+            pass
+
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        user = serializer.save()
+
+        # creating token
```
RobertBeaudenon, line 50: You can remove the creating of token now because the user that is created is not being redirected or something.

server/user_account/views.py
```diff
@@ -6,14 +6,59 @@
...
+        signals.post_save.send(sender=self.__class__,
+                               user=user, request=self.request)
+        token = Token.objects.get(user=user).key
+
+        data = {'user': user.user_name, 'token': token}
+        return Response(data, status=status.HTTP_201_CREATED)
+
+class OpenRegistrationView(viewsets.ModelViewSet):
+
```
RobertBeaudenon, line 59: Good that you kept it so we can test :)

RobertBeaudenon (changes requested):
Looks good. This will be merged soon when the minor changes will be addressed.

server/user_account/views.py
makchamp: That's true it is already handled by the permission class. I removed in dade84b836c4f152d1802526d2116771e057f97a

server/user_account/tests.py
makchamp: I added a detailed test case 5246f229da4dca5117c5e85036a64cd0d41ecd18
Only thing missing is handling the rules for who can register who. Once that is implemented, more test cases can be added.

RobertBeaudenon:
Looks good. Thank you for addressing the changes. I'll merge.

RobertBeaudenon (approved):

PR #41, opened 2020-10-14
Client gridview

Averynder:
This PR generates a table based on the client information for displaying the clients in the database. It has been structured in a way that allows it to be re-useable for other components who wish to display this table.


NicholasNagy (changes requested):
I added some review comments, you should additionally add the search function like we had discussed

client/src/app/components/client-gridview/client-gridview.component.html
```diff
@@ -0,0 +1,18 @@
...
+        <th>Is Active</th>
+        <th>E-mail</th>
+    </tr>
+    <tr *ngFor="let pers of users">
+        <td>{{pers.user_name}}</td>
+        <td>{{pers.first_name}}</td>
+        <td>{{pers.last_name}}</td>
+        <td>{{pers.role}}</td>
+        <td>{{pers.is_active}}</td>
+        <td>{{pers.email}}</td>
```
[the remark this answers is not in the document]
RobertBeaudenon: And this will be handled in the backend by returning to the frontend only the required fields ( so for this purpose in the backend we would have a specific serializer for the view that just returns the fields that we want)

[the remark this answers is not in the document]
NicholasNagy: if anything for now, let's not hard code the user's values. 
Just iterate over the user's values so that it can be dynamic for whatever fields are returned by the API for a user

client/src/app/components/modify-members/modify-members.component.ts
```diff
@@ -1,29 +1,27 @@
...
-import { User } from 'src/app/models/user.model';
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { ClientGridviewComponent } from 'src/app/components/client-gridview/client-gridview.component';
 import { ManageMembersService } from 'src/app/services/manage-members.service';
 
 @Component({
   selector: 'app-modify-members',
-  templateUrl: './modify-members.component.html',
-  styleUrls: ['./modify-members.component.css']
+  templateUrl: 'modify-members.component.html', // ../client-gridview/client-gridview
```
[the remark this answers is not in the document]
Averynder: fixed in commit a3e471

Averynder:
> 
> 
> I added some review comments, you should additionally add the search function like we had discussed

you would like that added to the current PR or have it as its own separate one?

client/src/app/components/client-gridview/client-gridview.component.html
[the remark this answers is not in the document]
Averynder: handled in the latest commit a3e471f6234310f1747af74339cb06ddb48ebe1d

server/user_account/views.py
```diff
@@ -110,12 +110,10 @@ def remove_token(self, request): # pylint: disable=no-self-use
...
 class AccessAllClients(generics.GenericAPIView):
     queryset = CustomUser.objects.values(
-        'user_name',
         'first_name',
         'last_name',
         'role',
         'is_active',
-        'email',
         )
     http_method_names = ['get']
```
RobertBeaudenon, line 119: You should refactor that view:
1- Don't hard code the value that you want to pass through this endpoint inside the queryset, you should create a serializer for that.
2- No need to specify http_method_names = ['get'] because you are using a GenericAPIView and you are already specifying in your get method definition that the operation that will only be allowed is a get.
3-You must secure this endpoint( Please check the last video I did on authentication under Sprint2/ in the drive or read the doc here https://www.django-rest-framework.org/api-guide/permissions/ )
5- Also you might want to add pagination, for now its okay but later when we will have a huge db we dont want to return all the query at once. (optional for now)
4-For now keep this method but modify the changes mentioned above, also this will probably be deleted because the viewset responsible for registration already implement this functionality (it handles CRUD) so we will just rename the registrationViewSet to UserViewSet) just to explain the future vision.

server/user_account/views.py
```diff
@@ -110,12 +110,10 @@ def remove_token(self, request): # pylint: disable=no-self-use
...
```
RobertBeaudenon, line 118: You should refactor that view:
1- Don't hard code the value that you want to pass through this endpoint inside the queryset, you should create a serializer for that.
2- No need to specify http_method_names = ['get'] because you are using a GenericAPIView and you are already specifying in your get method definition that the operation that will only be allowed is a get.
3-You must secure this endpoint( Please check the last video I did on authentication under Sprint2/ in the drive or read the doc here https://www.django-rest-framework.org/api-guide/permissions/ )
5- Also you might want to add pagination, for now its okay but later when we will have a huge db we dont want to return all the query at once. (optional for now)
4-For now keep this method but modify the changes mentioned above, also this will probably be deleted because the viewset responsible for registration already implement this functionality (it handles CRUD) so we will just rename the registrationViewSet to UserViewSet) just to explain the future vision.

client/src/app/components/client-gridview/client-gridview.component.html
```diff
@@ -0,0 +1,13 @@
...
+        <th>Is Active</th>
+        <th>Last Name</th>
+        <th>Role</th>
+    </tr>
+    <tr *ngFor="let item of users">
+        <td *ngFor="let item1 of item | keyvalue">
+            {{item1.value}}
+        </td>
+    </tr>
+</table>
```
RobertBeaudenon, line 13: If you are working on the frontend, you should use angular material. In this particular case, there is a component for tables ( https://material.angular.io/components/table/overview ) please implement it. I'm asking that because @makchamp already did the basic functionality to see the user in the frontend so redoing that without making it pretty is not really relevant because it will need to be refactored later (extra work).

client/src/app/components/modify-members/modify-members.component.html
```diff
@@ -1,12 +1,3 @@
...
-        username = {{pers.user_name}},
-        firstname = {{pers.first_name}},
-        lastname = {{pers.last_name}}, 
-        role = {{pers.role}},
-        is_active = {{pers.is_active}},
-        e-mail = {{pers.email}}
-    </li>
-</ul>
\ No newline at end of file
+<app-client-gridview></app-client-gridview>
```
RobertBeaudenon, line 3: So here you are just injecting the table client-gridview. Now what I'm suggesting is inside the client-gridview component you create 2 extra columns for the table and add the update and delete buttons that will be attached to the id of the particular user in that row. Also those 2 column will only be displayed if the logged-in user has the particular clearance and this can be handled using a ngIf*.

client/src/app/models/user.model.ts
```diff
@@ -1,8 +1,6 @@
 export interface User {
```
RobertBeaudenon, line 1: don't modify the interface we might this later, you can just create multiple interfaces for the different forms of the user that we are returning from the backend.

RobertBeaudenon (changes requested):
Please address the comments I just added.

Averynder:
> 
> 
> I added some review comments, you should additionally add the search function like we had discussed

Added in commit b9d15ea4438210143bedf032d85ad6b55e2ce729

client/src/app/components/client-gridview/client-gridview.component.html
Averynder: I'm not sure what you're referring too, I first coded the obtaining of the client information in this commit: 0ce7e2150cf1628464a80412fe080f2efbb3231a from a previous PR and then followed up in the current PR by creating a basic table in commit 6594d4b71879ebbb244c23ebc4854382134914f9 . Please look at the commits before commenting. As of commit ef691e7f16ed2ed697ba24007f9fd40a58b457eb angular material has been used

client/src/app/components/modify-members/modify-members.component.html
Averynder: This does not refer to the current PR. The goal of this PR was simply to obtain the already gotten clients and make a simple table out of it with addition to @NicholasNagy 's request of adding a search feature. I do not know if this is the smartest way or if we should go with an input text boxes along with the previous data as a placeholder. These are different ways and I believe this should be brought to it's own issue, it's own PR at the very least and depends on the mockups which have not yet been created for this

client/src/app/models/user.model.ts
Averynder: As per @NicholasNagy 's request: "We should probably not be exposing anything other than the first name, last name, the role and whether they are active" I believe that leaving sensitive information from a request could lead to security issues. This comment refers to future possibilities which isn't the goal of the PR, I would ask to once again remember that PR is meant for displaying the client information in a table, having it be re-useable for other components and include a search button

Averynder:
> Please use angular material

Please keep responses and feedback to actual reviews rather then as per commits. As visible above this is a work in progress and the goal is to make the commits atomic meaning they add a specific functionality and if too big are seperated rather then combining all requests into one. I will get to the feedback but as visible in my reply above this comment was placed twice but I have committed the changes for this request in commit ef691e7f16ed2ed697ba24007f9fd40a58b457eb . This simply leads to many duplicate responses. In order to avoid confusion, I will convert this PR into a draft and request a review when the all the requested changes have been accomplished and then we can continue to build upon this PR to add the desired functionality to production. Thank you, I appreciate your time

server/user_account/views.py
Averynder: 2) Addressed in commit b9d15ea4438210143bedf032d85ad6b55e2ce729

Averynder: 5- Addressed in commit ef691e7f16ed2ed697ba24007f9fd40a58b457eb

Averynder: 4- I have no idea why this is being commented, please keep review comments to requested changes

RobertBeaudenon: 4- Since you worked on this method and you did a change on it i’m just explaining the future vision of that since i was adding comments on it.

client/src/app/components/client-gridview/client-gridview.component.html
RobertBeaudenon: I looked at the commits before commenting, you should probably look at the comments i add before doing changes, because at first you just had a simple table that was not following the convention.

Averynder:
> Please watch again the videos, you should never do API calls inside Components it should be done inside services.
1) Please specify which videos
2) Fixed anyways in commit b4d9401fc28332bc24fae93483d1f3e5115b6d35

client/src/app/components/client-gridview/client-gridview.component.html
Averynder: > 
> 
> I looked at the commits before commenting, you should probably look at the comments i add before doing changes, because at first you just had a simple table that was not following the convention.

That was literally the point of this PR and **that I made**, not @makchamp , making it alone relevant and making the refactoring of it unnecessary as per your first comment, however as per your request this has been resolved

server/user_account/views.py
Averynder: 1. Fixed as of commit d3e6a3ed4f19148693879d75a0b8c0de80072ddb

Averynder: 3. Fixed as of commit: beaa57e70e4afaf1bee491c6f27c0196e85ca8ed

client/src/app/components/client-gridview/client-gridview.component.spec.ts
```diff
@@ -0,0 +1 @@
+
```
RobertBeaudenon, line 1: Please you need to test the initialization of the component.

client/src/app/components/client-gridview/client-gridview.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  applyFilter(event: Event): void {
+
+    if (this.dataSource.paginator) {
+      this.dataSource.paginator.firstPage();
+    }
+  }
+
+  updateClients(clients): void {
```
RobertBeaudenon, line 46: Please rename this function to 'populateTable(clients)' or something similar to that. The update keyword might confuse others that you are actually modifying infosof users.

client/src/app/components/client-gridview/client-gridview.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  @ViewChild(MatPaginator) paginator: MatPaginator;
+
+  constructor(
+    private manageMembersService: ManageMembersService,
+    private changeDetectorRefs: ChangeDetectorRef) {
+    this.users = new Array<User>();
+
+    // Assign the data to the data source for the table to render
+    this.manageMembersService.getAllClients()
```
RobertBeaudenon, line 28: I would say that you can move this block inside a function called init() or something around that because in this case, you are initializing the table by getting all the clients. Then inside your ngOnInit() method you can call this init() method so that when a component will be created it will do all of those actions you are doing it (even though what you did is working ). Also, the ngOnInit() should be the first method after the constructor, so please reorder them.

client/src/app/components/modify-members/modify-members.component.html
```diff
@@ -1,12 +1,14 @@
...
-    </li>
-</ul>
\ No newline at end of file
+<form [formGroup]="checkoutForm" (ngSubmit)="onSubmit(checkoutForm.value)">
+
+  <div>
+    <label for="name">
+      FirstName
+    </label>
+    <input id="name" type="text" formControlName="name">
```
RobertBeaudenon, line 8: Please add angular material for the Input, the label and the button. I believe all of those are already imported in the matmodule so it's quick. Just each time you work on the html add angular material stuff.

client/src/app/services/manage-members.service.ts
```diff
@@ -15,4 +15,9 @@ export class ManageMembersService {
   getAllClients(): Observable<any> {
     return this.http.get<User[]>(`${this.BASEURL}/getAllClients/`);
   }
+
+  sendNameToObtainClients(name): Observable<any>
```
RobertBeaudenon, line 19: I feel the method name is a bit weird, you can just call it getSpecificClients(name) or something around that.

client/src/app/services/manage-members.service.ts
```diff
@@ -15,4 +15,9 @@ export class ManageMembersService {
...
+  {
+    return this.http.post('http://localhost:8000/getSomeClients/', JSON.stringify(name));
```
RobertBeaudenon, line 21: Don't hard code the URL inside the request, create a BASEURL variable on top where you specify the global URL used. you can look at how it's done inside the other services

server/user_account/views.py
```diff
@@ -145,18 +142,29 @@ def remove_token(self, request): # pylint: disable=no-self-use
...
-        'last_name',
-        'role',
-        'is_active',
-        'email',
-        )
-    http_method_names = ['get']
-
+class AccessAllClients(viewsets.ModelViewSet):
+    @api_view(['Get'])
+    @permission_classes([IsAuthenticated])
```
RobertBeaudenon, line 147: Statements at line 146 and 147 should not be used in class based views, it is only used in function based views so it doesn't make sense to have it here. If you are a bit confused at the difference between class based views and function based views the tutorials on Django that @NicholasNagy and I did explains it really well (Or again you have the documentation.

server/user_account/views.py
```diff
@@ -145,18 +142,29 @@ def remove_token(self, request): # pylint: disable=no-self-use
...
     def get(self, request):
```
RobertBeaudenon, line 148: Also I'm not even sure how this function is working because you are declaring your class based view as a modelViewSet and inside of it you are creating a function based view which does not make sense. I know the concept about class based views and function based views are confusing, If you want we can get on a call and I"ll be glad to help you clariy some stuff.

server/user_account/views.py
```diff
@@ -145,18 +142,29 @@ def remove_token(self, request): # pylint: disable=no-self-use
...
-        qs_json = json.dumps(list(qs), cls=DjangoJSONEncoder)
-        return JsonResponse(qs_json, safe=False)
+        instance = self.get_object()
+        serializer = self.get_serializer(instance)
+        serializer.is_valid(raise_exception=True)
+        return Response(serializer.data)
+
+    permission_classes = [IsAuthenticated]
+    queryset = CustomUser.objects.all()
+    serializer_class = ClientGridSerializer
```
RobertBeaudenon, line 156: Again this should be placed right after the declaration o the class-based view.

server/user_account/views.py
```diff
@@ -145,18 +142,29 @@ def remove_token(self, request): # pylint: disable=no-self-use
...
-        'user_name',
-        'first_name',
-        'last_name',
-        'role',
-        'is_active',
-        'email',
-        )
-    http_method_names = ['get']
-
+class AccessAllClients(viewsets.ModelViewSet):
```
RobertBeaudenon, line 145: Is this even being called? I don't see any URL redirecting to it. If not used delete it please.

RobertBeaudenon (changes requested):
Regarding the frontend you forgot to add a test and i requested some minor modifications.
Regarding the backend I added some comments to explain some of your errors.
Once the changes will be fixed I'll merge.

client/src/app/components/client-gridview/client-gridview.component.spec.ts
Averynder: fixed in commit 784cde4613b4b1f0bc0728d793ac2f923919a2a9

client/src/app/components/client-gridview/client-gridview.component.ts
Averynder: > Please rename this function to 'populateTable(clients)' or something similar to that. The update keyword might confuse others that you are actually modifying infosof users.

Fixed in commit 38fa446370aa1c6aa2e3d4df0f78677c0e8076b2

client/src/app/components/client-gridview/client-gridview.component.ts
Averynder: Fixed in commit 38fa446370aa1c6aa2e3d4df0f78677c0e8076b2

client/src/app/services/manage-members.service.ts
Averynder: Fixed in commit 38fa446370aa1c6aa2e3d4df0f78677c0e8076b2

client/src/app/services/manage-members.service.ts
Averynder: fixed in commit 38fa446370aa1c6aa2e3d4df0f78677c0e8076b2

client/src/app/components/modify-members/modify-members.component.html
Averynder: Fixed in commit f746527c72e44fe8b5f91554b127d5315ab3fd52

RobertBeaudenon (approved):

NicholasNagy (approved):

Issue #29, opened 2020-10-06
Set up Code Coverage


NicholasNagy:
Set up Code Coverage for our testing tools.

They must be able to run and give test coverage results for all tests (unit and system) for both the front-end and back-end.

PR #68, opened 2020-10-18
Issue #29 Set up Code Coverage

Neeham:
This PR is about Quality assurance for this project.

**Code Coverage**

- Frontend (Angular): Using ng test with Karma. The coverage can be seen under CI console. 

- Backend (Django): Using pytest-cov to obtain coverage and exporting the result as XML which is then pushed to CodeCov. Results can be seen at the following URL: https://codecov.io/gh/Fruity-Loops/ALTA

Threshold for Backend: Overall of 70% coverage is set as the threshold. If the PR being merged with master causes a drop of more than 10% coverage, this will be reported as a failure.

**SonarCloud**

- Frontend (Angular): Since the main focus is on the Backend, for now, I have excluded this from SonarCloud. This can be added if needed but will have to discuss about the paths to exclude from the **client/** directory that should not be considered under SonarCloud. 

- Backend (Django): Everything under **server/** directory is pushed to SonarCloud, if any exclusion is needed this can be added under **sonar-project.properties** found at the root directory of this repo. Results can be seen at the following URL: https://sonarcloud.io/dashboard?id=Fruity-Loops_ALTA

### DIFF
--- a/.github/workflows/client_workflow.yml
+++ b/.github/workflows/client_workflow.yml
@@ -26,4 +26,4 @@ jobs:
           npm i
           npm i -g @angular/cli
           npm audit fix
-          ng test --browsers=ChromeHeadless --sourceMap=false --watch=false 
+          ng test --no-watch --code-coverage --browsers=ChromeHeadless --sourceMap=false --watch=false
--- a/.github/workflows/server_workflow.yml
+++ b/.github/workflows/server_workflow.yml
@@ -27,17 +27,16 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-    - name: Running tests
+    - name: Makemigrations and Migrate
       working-directory: server
       run: |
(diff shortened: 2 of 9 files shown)


NicholasNagy (changes requested):
Everything looks good, but you should rebase and rename your commits based on the committing convention we have.

Also, when I look on the branch, it's not displaying the code coverage %, and I was wondering why?

Neeham:
> Everything looks good, but you should rebase and rename your commits based on the committing convention we have.

Fixed 😄 

> Also, when I look on the branch, it's not displaying the code coverage %, and I was wondering why?

This is because it's set to show it on the master branch. Once this PR is accepted, the master branch will show the coverage % and then for future PR it will refer to master branch to compare and see how much up or down the coverage is going.

NicholasNagy (approved):
Cool, everything looks good now, so I'll merge

Issue #71, opened 2020-10-18
User Story 1.3A


NicholasNagy:
### User Story: ALTA-1.3A
As a system administrator, I want to be able to add a new organization to ALTA so that I can build my clientele.

### Description

* Required data to create an organization:
  * Name

- Not required data:
  - Locations

* Upon creation, the organization's account status is set to _Active_.

- Possible organization types:
  - Not specified - no locations 
  - Specific - contains 1 location
  - Global - contains 2+ locations

### Acceptance Criteria (AT-1.3A #133)
Given that the user is logged in as a system administrator,

1. The user should be redirected to the _Manage Organizations_ page which displays a table containing all organizations subscribed to ALTA.
2. The user selects the  _Add_ button.
3. The user enters the organization's name into the popup input field.
4. The user submits the new organization by selecting the _Create_ button.

### Related Mockups

- System administrator logs into ALTA
<img width="465" alt="Capture d’écran 2021-01-03 à 12 01 51" src="https://user-images.githubusercontent.com/35980232/103477100-81fd0d00-4dbb-11eb-878d-f3da937a321f.png">

- Table of subscribed organizations
<img width="465" alt="Capture d’écran 2021-01-04 à 17 43 38" src="https://user-images.githubusercontent.com/35980232/103558099-72152400-4eb4-11eb-8a8a-7c89a7777ceb.png">

- Create organization form
 > Can either input a single location (by default will be considered as a Specific location), or many by uploading an .xlsx file (will be considered as global)
<img width="465" alt="organization form" src="https://user-images.githubusercontent.com/35980232/103808185-db816800-5057-11eb-81f6-efe88676ae5c.png">

<img width="465" alt="organization submit" src="https://user-images.githubusercontent.com/35980232/103809016-2780dc80-5059-11eb-95d4-b57c8ab5d105.png">

PR #74, opened 2020-10-20
Issue #71 Manage organizations backend

RobertBeaudenon:
User Story 1.6
Provided the following functionalities:
- Created a model responsible of holding organization information
- Created an API endpoint that allows user to perform CRUD operations on organizations
- Added custom permissions on the API endpoint
  - A System Admin is allowed  to: Create, Update, Delete, Retrieve and Organization
  - An  Inventory Manager is allowed to: Update and Retrieve and Organization
- Provided test cases for that


server/django_server/permissions.py
```diff
@@ -0,0 +1,38 @@
...
+        This can be used either inside the permission_class directly or you can call it from other permission files
+        :param request:
+        :param view:
+        :return: True/False : Whether the user is a Inventory Manager or Not
+        """
+        user = CustomUser.objects.get(user_name=request.user)
+        if user.role == 'IM':
+            return True
+        else:
+            return False
```
makchamp, line 38: You can return the expression itself to save loc
return user.role == 'IM'

server/organization/permissions.py
```diff
@@ -0,0 +1,20 @@
...
+             :param request:
+             :param view:
+             :return: True/False : Whether the user is allowed to perform CRUD
+             """
+        if view.action in ['list', 'retrieve', 'update', 'partial_update']:
+            return isSystemAdmin.has_permission(None, request, None) or isInventoryManager.has_permission(None, request, None)
+        elif view.action in ['create', 'destroy']:
+            return isSystemAdmin.has_permission(None, request, None)
+        else:
+            return False
```
makchamp, line 20: same here you can return the expression itself

server/organization/tests.py
```diff
@@ -0,0 +1,60 @@
...
+from django.test import TestCase
+from django.test.client import Client as HttpClient
+from rest_framework import status
+from rest_framework.test import APITestCase
+from rest_framework.authtoken.models import Token
+from rest_framework.test import APIClient
+from user_account.models import CustomUser
+
+
+class OpenRegistrationTestCase(APITestCase):
```
makchamp, line 11: This test case is dummy for an endpoint that will not be used in production. Maybe add these tests to the actual Registration endpoint test case class or create a new one just for organization ?

makchamp (changes requested):
Good change. Will merge once the errors are cleaned up.
However I would like to raise one point. I did not see how you are storing which user belongs to which organization. An option would be to give the user model a reference to the organization model (through a foreign key or something). Maybe we should reference our design docs/diagrams for this.

server/django_server/permissions.py
RobertBeaudenon: Yes, I'll fix it.

server/organization/permissions.py
RobertBeaudenon: In this case, i cant since im using elif.

server/organization/tests.py
RobertBeaudenon: it's for organization I just copied the structure but forgot to change the name of the class hahaha

makchamp (approved):
Thanks @RobertBeaudenon for implementing the requested changes.
As for linking the user and organization models, it was discussed that further work on the matter will be done in a future PR.

Issue #39, opened 2020-10-12
User Story 1.5


Amanda-Kolopa:
As a system administrator, I want to be able to modify a client’s employee’s user account so that I can reflect any changes requested to their account information.

### Description:

Allow system admins to go into an organization, access a list of employees, and be able to select any employee to be able to modify their info.

### Acceptance Test (AT-1.5 #135) :

Given that the user is logged in as a system administrator and has selected an organization from the Manage Organizations page,

The user selects the Employees side menu option to present a table containing the affiliated employee user accounts' information.

The user selects the grey dot button at the end of the respective employee's row to present the form containing the employee's account information.

The user selects the Edit button to allow the fields to accept input.

Once the information is modified, the user selects the Save button to store the updated data. The form is then refreshed to display the updated information.

### Related Mockups (to be updated*):

- For logging in
![image](https://user-images.githubusercontent.com/40705372/101196779-2d853700-362f-11eb-9d77-0b4545669d3a.png)

- For listing out the organizations
![image](https://user-images.githubusercontent.com/40705372/101197185-c6b44d80-362f-11eb-85a3-e8a723c17346.png)

- For listing out the employees*
![image](https://user-images.githubusercontent.com/40705372/101196867-50175000-362f-11eb-8597-b884dda8006d.png)

- The editing page for an employee*
![image](https://user-images.githubusercontent.com/40705372/101196921-658c7a00-362f-11eb-9904-e2560b444435.png)

PR #75, opened 2020-10-21
Modify User attributes #39

Averynder:
This uses the previous PRs:
1) where client information was obtained from the database and displayed
2) the information was made into a proper table format

and adds the following functionality
1) The user / System admin can now modify any of the fields and they will be reflected in the table
2) The changes will also be reflected in the database

This will close issue #39


client/src/app/components/client-gridview/client-gridview.component.html
```diff
@@ -4,25 +4,33 @@
...
+        </td>
       </ng-container>
   
       <!-- Color Column -->
       <ng-container matColumnDef="Role">
         <th mat-header-cell *matHeaderCellDef mat-sort-header> Role </th>
-        <td mat-cell *matCellDef="let row" > {{row.role}} </td>
+        <td mat-cell *matCellDef="let row" >
+          <span (keyup)="changeValue(row.id, 'role', $event)" (blur)="updateList(row.id, 'role', $event)" contenteditable="true">{{row.role}}</span>
+        </td>
```
RobertBeaudenon, line 33: Nice effort to make those fields editable, I can modify rows (with errors). In my opinion, this is not the best way to do it and I'll explain why. First, If I make errors while modifying those rows meaning that I add a value that is not allowed for example for a role i'm only allowed to be an SA, IM or SK, but you are allowing to modify any value there without throwing any error. Second, this is also happening for the true/flase Isactive column, it will change the value but this time if you refresh its throwing an error inside the console so it resets it. So my suggestion to make this thing better in terms of user experience, is that i would forget about that idea of modifying directly the table because this can lead to confusion and if as a developer you want to throw an error if fields are wrongly modified how would you do this in a clean way. So what I did in another project where I had to modify table rows that you could do is add a button on the right of each row called to update or you can simply add a click listener on the row, once clicked it will pop out a view card ( for example like the one I did for registration) that just holds the info of the row you clicked with the fields prepopulated and there you will be able to handle the errors more properly and for the fields that require specific options like the role or isactive you can add a dropdown menu for the list of choices.

RobertBeaudenon: Look at the inconsistencies that i was able to create inside Is Active and Role:
Screen Shot 2020-10-21 at 6 13 53 PM

client/src/app/components/client-gridview/client-gridview.component.ts
```diff
@@ -55,4 +56,16 @@ export class ClientGridviewComponent implements OnInit {
       this.changeDetectorRefs.detectChanges();
   }
 
+  updateList(id: number, property: string, event: any): void {
+    const editField = event.target.textContent;
+    this.users[id][property] = editField;
+    this.manageMembersService.modifyClientInfo(property, editField, id).subscribe((user) => {
+      const users = user;
+      this.populateTable(users);
+    });
```
RobertBeaudenon, line 65: You should handle errors that are sent back by the API. Check out how it's done in other components.

client/src/app/components/client-gridview/client-gridview.component.ts
```diff
@@ -55,4 +56,16 @@ export class ClientGridviewComponent implements OnInit {
       this.changeDetectorRefs.detectChanges();
   }
 
+  updateList(id: number, property: string, event: any): void {
```
RobertBeaudenon, line 59: You can rename this method to updateUser() or something around that. Please also from now on i think its a good idea to add a comment on what the method is doing just on top of it or whatever is the convention so that when others will start implementing it will be obvious what is doing what.

client/src/app/services/manage-members.service.ts
```diff
@@ -20,4 +20,9 @@ export class ManageMembersService {
   {
     return this.http.post(`${this.BASEURL}/getSomeClients/`, JSON.stringify(name));
   }
+
+  modifyClientInfo(category, field, id): Observable<any>
+  {
+    return this.http.post(`${this.BASEURL}/modifyClients/`, JSON.stringify({'category': category, 'field': field, 'id': id}));
```
RobertBeaudenon, line 26: Instead of doing "JSON.stringify({'category': category, 'field': field, 'id': id})" you can just create an object called 'body' and add all those properties to it and then pass the body instead of that statement.

server/user_account/urls.py
```diff
@@ -15,5 +15,6 @@
     path('', include(router.urls)),
     path('login/', views.LoginView.as_view()),
     path('logout/', views.LogoutView.as_view()),
-    path('getSomeClients/', views.AccessSomeClients.as_view())
+    path('getSomeClients/', views.AccessSomeClients.as_view()),
+    path('modifyClients/', views.ModifyClient.as_view()),
```
RobertBeaudenon, line 19: Minor but I would rename this endpoint to modifyClient/ instead of modifyClients/ because you are modifying one client at a time.

server/user_account/views.py
```diff
@@ -161,3 +160,23 @@ def post(self, request):
...
+        variable_column = data.get('category', '')
+        entry = data.get('field', '')
+        id1 = data.get('id', '')
+        if entry == 'false':
+            entry = False
+        elif entry == 'true':
+            entry = True
+        CustomUser.objects.filter(id=id1).update(**{variable_column: entry})
+        serializer = ClientGridSerializer(instance=self.get_queryset(), many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
RobertBeaudenon, line 182: For the moment this method is okay by you can simplify all of what is inside the post method by using a viewset. If you want you can checkout what i did in my branch (manage_organization_backend) which is now a pull request. And if you decide to convert it to a viewset you can just combine it with the other view you created class AccessAllClients(viewsets.ModelViewSet) that will have both methods under the hood. If you need I can explain on a call more about the power fo viewsets, let me know.

RobertBeaudenon (changes requested):
Nice effort by implementing the modification of clients in the table, but you are not handeling the errors ( no restrictions on user role, you can add whatever value and it will be validated).
- I suggested a better way of making the modification cleaner inside a card that will pop out on top of the table.
- Regarding the backend it is working, but also can be refactored.

server/user_account/views.py
NicholasNagy: Also, on a side-note, this method isn't tested.

client/src/app/components/client-gridview/client-gridview.component.ts
```diff
@@ -55,4 +67,36 @@ export class ClientGridviewComponent implements OnInit {
...
+      }
+
+      // If name contains illegal characters
+      if (err.error.first_name) {
+        this.errorMessage = err.error.first_name[0];
+      }
+    });
+  }
+
+  changeValue(id: number, property: string, event: any): void {
```
RobertBeaudenon, line 99: Why do you still have the id and property arguments of the function if you are not using them.

client/src/app/components/client-gridview/client-gridview.component.ts
```diff
@@ -16,6 +16,18 @@ export class ClientGridviewComponent implements OnInit {
   dataSource: MatTableDataSource<User>;
```
RobertBeaudenon, line 16: The imports are not shown here but the AfterViewInit import is not being used so please delete it.

client/src/app/components/client-gridview/client-gridview.component.html
```diff
@@ -1,28 +1,44 @@
 <div class="mat-elevation-z8">
     <table mat-table [dataSource]="dataSource" matSort>
```
RobertBeaudenon, line 3: Although it's not my preferred way of modifying the table directly you made it work but there is still one problem here. When I modify one row of  First name one it is modified for some reason it is modifying the one that is under it also so I think you might have a problem with using two times the same ID or something like that. This also applies to IsActive and Lastname.

RobertBeaudenon (changes requested):
Things are way more cleaner now thanks for taking into consideration the comments previously added. I wanted to mention, some of the mistakes that I am adding a comment on can be avoided before you commit to the pull request such as unused imports, not used arguments in methods, or commented lines of code that forgot to be deleted and also you should add comments to appropriate views. Please next time make sure to verify those because it will make my job easier. 
I requested little changes before merging:
- Please merge whatever was pushed on master branch
- There is a bug with the table
-  Add comments to views in the backend
- Also look carefully on whatever other comments i added

NicholasNagy:
I will fix lint and add tests tomorrow, but for now should be good to look at for feedback @RobertBeaudenon ,
lmk especially in regards to the form.

server/user_account/views.py
```diff
@@ -1,13 +1,39 @@
...
+        employee_serializer = ClientGridSerializer(employee)
+        return Response(employee_serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        employee_serializer = ClientGridSerializer(employee, data=request.data)
+        if employee_serializer.is_valid():
+            employee_serializer.save()
+            return Response(employee_serializer.data, status=status.HTTP_200_OK)
+        return Response(employee_serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
```
RobertBeaudenon, line 35: Just to explain the future vision: The "GET" a specific employee would now be done through the RegistrationView which is now renamed to CustomUserView. The "PUT" will now be done throught the View that I just created called UpdateProfileView, although I will just add permissions to these two views to be not only accessible by the owner but also by the admin.

server/user_account/serializers.py
```diff
@@ -43,4 +43,4 @@ class Meta:
 class ClientGridSerializer(serializers.ModelSerializer):
     class Meta:
         model = CustomUser
-        fields = ['first_name', 'last_name', 'role', 'is_active']
+        fields = ['first_name', 'last_name', 'role', 'is_active', 'id', 'email']
```
RobertBeaudenon, line 46: So this will be modified because we don't want id and email to be returned for all the users in the table, We just want this extra info when we 'get' a specific user. But I know that this is temporary, and will be done through the UserSerializer.

client/src/app/modules/alta-main-routing/alta-main-routing.module.ts
```diff
@@ -21,8 +22,9 @@ const routes: Routes = [
       },
       { path: 'dashboard', component: DashboardComponent },
       { path: 'create-members', component: CreateMembersComponent },
-      { path: 'modify-members', component: ModifyMembersComponent },
-      {path: 'manage-organizations', component: ManageOrganizationsComponent}
+      { path: 'modify-members', component: ClientGridviewComponent },
+      {path: 'manage-organizations', component: ManageOrganizationsComponent},
+      {path: 'modify-members/:ID', component: EmployeeSettingsComponent}
```
RobertBeaudenon, line 27: Good.

client/src/app/components/employee-settings/employee-settings.component.ts
```diff
@@ -0,0 +1,82 @@
...
+import {ActivatedRoute} from '@angular/router';
+import {User} from '../../models/user.model';
+
+@Component({
+  selector: 'app-employee-settings',
+  templateUrl: './employee-settings.component.html',
+  styleUrls: ['./employee-settings.component.scss']
+})
+export class EmployeeSettingsComponent implements OnInit {
+  @Input() employee: User;
```
RobertBeaudenon, line 12: So instead of injecting from parent to child ( because later i think we wont have all the informations of the user at once in the table), I think we could just do a api call just to get the info of this particular user, this way the component would be reusable for example for the update profile of the logged in user.

client/src/app/components/employee-settings/employee-settings.component.html
```diff
@@ -0,0 +1,59 @@
...
+          </mat-form-field>
+
+          <mat-form-field>
+            <mat-label for="email">Email</mat-label>
+            <input matInput id="email" type="text" [(ngModel)]="employee.email" email [disabled]="!edit"/>
+          </mat-form-field>
+
+          <mat-form-field>
+            <mat-label for="password">Password</mat-label>
+            <input matInput id="password" type="password" [(ngModel)]="password" minlength="5" [disabled]="!edit"/>
```
RobertBeaudenon, line 31: I think the password should be moved to another form just under it with 2 inputs (new password, confirm password), and this will be updated through a specific endpoint that is just responsible for updating the password

client/src/app/components/employee-settings/employee-settings.component.html
```diff
@@ -0,0 +1,59 @@
...
+        <mat-card-title>
+          {{edit? "Update": ""}} {{employee?.first_name + " " + employee?.last_name}}
+        </mat-card-title>
+      </mat-card-header>
+
+      <mat-card-content>
+
+          <mat-form-field>
+            <mat-label for="First Name"></mat-label>
+            <input matInput id="first_name" type="text" [disabled]="true" [value]="employee.first_name"/>
```
RobertBeaudenon, line 16: Why can't we update the first name

client/src/app/components/employee-settings/employee-settings.component.html
```diff
@@ -0,0 +1,59 @@
...
+          </mat-form-field>
+
+          <mat-form-field>
+            <mat-label for="Last Name"></mat-label>
+            <input matInput id="last_name" type="text" [disabled]="true" [value]="employee.last_name"/>
```
RobertBeaudenon, line 21: why can't we update the last name

RobertBeaudenon (commented):
Overall looks good, I added some comments just to explain how I think it can be improved to make the employee settings component reusable for updating the profile of the current logged in user. Let me know your opinion on what i mentioned, please.

client/src/app/components/employee-settings/employee-settings.component.html
NicholasNagy: People's names don't change that easily

server/user_account/serializers.py
NicholasNagy: yep exactly, and I think this entire class was used exactly for that, so we're probably going to have to refactor and remove this whole class anyway

server/user_account/views.py
NicholasNagy: Why do are we splitting up the views like that?
Wouldn't it just make more sense to have them all under a single class for CRUD operations?

client/src/app/components/employee-settings/employee-settings.component.ts
NicholasNagy: I don't understand what you mean. 
What I have implemented is on generation of this component, it will make an API call to get the user information.
The component is entirely reusable, the only consideration we'd have to make is how we pass the ID

NicholasNagy: Since right now, it's passed through the url.
What we could do is just make a route for the settings to pass the ID, which could also work as well, and might make more sense.
I don't know if there's a cleaner way to do this. In React, I would just pass the ID down to the component, and display only that component. However in Angular, it kind of feels cleaner to just display the new page through a different route.

client/src/app/components/employee-settings/employee-settings.component.html
NicholasNagy: So right now, it's actually not associated to any form.
This was something that I wanted to discuss as well, because for the selection inputs (role and active), I couldn't find a viable way to set the default values on those fields. Using ngModel has been a cleaner alternative, but the consideration we have to make is how we validate the input.
This is something we can always refactor later if we want to add validation, but just some food for thought for now.

server/user_account/views.py
RobertBeaudenon: Well actually we could do that using the RetrieveUpdateAPIView as shown in the DRF doc:
Screen Shot 2020-10-31 at 3 20 09 PM

But what I did in the other pull request is that i just used an UpdateAPIView as shown in the DRF doc:
Screen Shot 2020-10-31 at 3 21 06 PM

because the "GET" was already provided by the viewset so If was like why not use it. Its just a matter of preferance. 

Also another reason why i chose to use the viewset for the GET is because the serializer is returning everything that we need for the user:
Screen Shot 2020-10-31 at 3 25 39 PM

So the Update serializer is more restricted as shown below:
Screen Shot 2020-10-31 at 3 26 37 PM

I won't allow the user to update any field that is not present here.

Let me know what do you think behind my reasoning.

client/src/app/components/employee-settings/employee-settings.component.ts
RobertBeaudenon: Oh sorry I misread, you are right i saw the API call in the ngOnInit(). Okay this is good then. Just another think instead of writing the whole function inside the ngOnInit() to get the particular user, i would just suggest to move it outside in another method called getEmployee() and then call this method inside ngOnInit().

RobertBeaudenon: I was getting the ID in my other projects through the URL too I think it's fine for now but for sure we can so some reaserch.

client/src/app/components/employee-settings/employee-settings.component.html
RobertBeaudenon: Yeah i know but what if they made a typo.

NicholasNagy: That would usually occur on account creation. If that's the case, they can always delete the user and create a new account.

client/src/app/components/employee-settings/employee-settings.component.ts
NicholasNagy: Hmm, well, I'm not sure how I could decompose it, right?
Like for example, I could refactor out the line of code from the service call + the service subscription functionality, but the service subscription functionality is fairly coupled with the initialization of the state of the component, so I feel like if I removed that part, it wouldn't make much sense.

Or if I decided to remove the line of code from the service call, I would still need to pass in a lambda function to initialize the state, which would end up looking like the same thing, but with an extra method, adding unnecessary code bloat.

In either of these cases, I'm not sure they are the best decisions. Do you have something particular in mind? 
Maybe I can add a comment to make it harder to miss that we are getting an Employee?

server/user_account/views.py
NicholasNagy: I see what you mean, but is there no way to set the serializers at a method level?
For example, 1 serializer can be used for get, and the next would be used for put and patch?

Also, another detail that I want to mention is that we should add more fields to the Update serializer. We should be able to modify email, role and is_active at the very least. We're going to be removing user_name, so that can be removed.

client/src/app/components/employee-settings/employee-settings.component.ts
RobertBeaudenon: I just meant like this:
Screen Shot 2020-10-31 at 4 19 12 PM

server/user_account/views.py
RobertBeaudenon: So for the first point, We could do that but since the serializer_class is usually generalized for all of the actions on a class based view the only way to define it is to override the specific method responsible for put/patch to say that we want to use another serializer class. So i thought about it because i could have implemented all of that inside the registration viewset but i thought it would be a bit a mess since i'm already overriding things there for the create user.
Do you understand my choice?

For the second point, yes i can add those but if i'm a user would i want to be able to change my role etc.. I'm talking about the case of any user updating their profile. That's why i didn't have those.

client/src/app/components/employee-settings/employee-settings.component.ts
NicholasNagy: Yeah, so that's my first point. Okay, I can do that though. 
I think it actually makes more sense now that I think about it, I was already initializing some variables in the component in other functions. I'll update this soon.

server/user_account/views.py
NicholasNagy: For the first point, I see why you did it, but have you tried this:
get_serializer

It should help you provide some flexibility to what serializers you return. You can maybe set it up to that based on the action, you will return a specific serializer.

For the second point, I see what you mean, but maybe also the above pic can help you also modify the serializer based on the role?

client/src/app/components/employee-settings/employee-settings.component.ts
NicholasNagy: Fixed!

server/user_account/views.py
RobertBeaudenon: I will look into the first point, but can you provide the link i'm not able to find it on the DRF doc.
For the second point, once i will look into it we can discuss on the other PR this choice.

NicholasNagy: Here's the link: [https://www.django-rest-framework.org/api-guide/generic-views/](https://www.django-rest-framework.org/api-guide/generic-views/)

It's in the generic-views part, but yeah let's take this over to PR #116

RobertBeaudenon:
Thanks for the good work, we can finally close this PR. I'm merging to master.

RobertBeaudenon (approved):

Issue #89, opened 2020-10-23
Mislabelled Create User Button


NicholasNagy:
### Details
**Summary** 
The label of the button for creating a new user is misrepresenting the actual functionality.

**Reproduction Steps:**

1. The user logs into the system and selects the _Create Members_ option from the side menu.
2. The page displays the empty registration form with a button to submit the new user data.

**Expected Results:**
The button displays the label “Save”. 
 

**Actual Results:**
The button displays the label “Sign up”. This intends that the logged-in user is creating a personal account rather than registering a new user.

### Tracking

**Severity:** Low
**Priority:** Medium
**Status:** Fixed

Issue #88, opened 2020-10-23
Registration Access for Everyone


NicholasNagy:
### Details
**Summary** 
The system requirements state that registration should only be performed by logged in users. The current setup permits the creation of accounts without logging into the system.

**Reproduction Steps:**

1. The user visits the ALTA landing page.

**Expected Results:**
The user is requested to log into the system with their email address and password.
 

**Actual Results:**
The user is presented with the option to log in or sign up.

### Tracking

**Severity:** Medium
**Priority:** High
**Status:** Fixed

PR #104, opened 2020-10-28
Bug fixing [Issue #88, #89] + Fix error message with signup form

makchamp:
#### Fix for Issue #88: Registration Access for Anyone
The signin tab now does not render at all if the 'open_registration' env variable is set to false.
There is no more tabs actually. Just the log in component on it's own (when the env variable is set to false).

#### Fix for Issue #89: Mislabelled Create User Button 
Renamed from 'Sign Up' to 'Register Account'
Also extracted the button label into a component variable for easier access or in case we would want to modify this label programmatically

#### Fix error message in signup form showing [Object] and it being shown by default
A call to the backend is being made to retrieve all organizations in the component's init function. This should probably be changed to only be called when the user actually clicks on the organization drop down menu. For now, a condition to make the request contingent on if there is an authenticated user logged in has been added. This also prevents the error box from displaying initially because the endpoint is protected anyway.
The error message inside the box was also not being displayed.

#### Renamed Auth-Tabs Component to Greeting Component 
The greeting page is the first page that is shown for any user looking to login/signup. The component is named as such.
The auth-tabs component is no longer a relevant name as the signup component will not be shown in the production build and because it is a less descriptive name for the page


NicholasNagy:
Are the last 2 fixes mentioned in your PR Description documented in our issues?

NicholasNagy (changes requested):
I have no comments related to the code, but we need to remove that organizations drop down. We're never going to need it, because system admins aren't associated with any organization, and when in the frame of an organization, we'll be able to access which organization that a user wants to make an account for.

NicholasNagy (approved):
Actually, I think that's more related to another user story, so forget it for now. 
I think in order to keep this PR concise, I'll open up another issue to fix that problem, since I think that this functionality is outside the scope of this PR.

PR #112, opened 2020-10-30
Migrated project from CSS to SASS

domludera:
Using [schematics-scss-migrate](https://www.npmjs.com/package/schematics-scss-migrate)
		- Converted .css files to .scss
		- Altered the styleUrls value to point to the new file names for the stylesheet

	Changed configuration such that subsequent components will generate .scss files


NicholasNagy (approved):
Looks good to me, I tried it locally and it compiles perfectly.

Thanks for this!

Issue #90, opened 2020-10-23
Dysfunctional Search


NicholasNagy:
### Details
**Summary** 
When searching through the list of registered users, the retrieved results are based on an exact match with the inputted username. The search functionality should be able to return the usernames which contain the given sequence of characters.

**Reproduction Steps:**

1. The user logs into the system and selects the _Manage Members_ option from the side menu.
2. The page displays a table listing all the users registered in the system and a search bar. The user queries the first few characters of their username. 

**Expected Results:**
At a minimum, the results display all users with a username that begins with the submitted characters. 

The search should be later improved to return results while the user inputs the values in the search bar.

**Actual Results:**
The user’s search only retrieves users with the exact match of the inputted value.

### Tracking

**Severity:** High
**Priority:** High
**Status:** Fixed

PR #114, opened 2020-10-30
Fix client search (issue #90) by adding auto-filtering (works with all fields)

makchamp:
- The search feature was dysfunctional as indicated in issue #90 
- I improved it by adding auto filtered results
- The search will filter by ALL fields automatically (first name, last name, role, etc..)
- I also increased the size of the search bar significantly 
- Increased the initial number of rows from 5 to 10


Averynder:
Works well, code is clean, just fix the lint issue for the typedef and this should be merged

Averynder (approved):

Issue #43, opened 2020-10-18
User Story 1.2


NicholasNagy:
### User Story: ALTA-1.2
As a system administrator, I want to be able to modify my account information so that I can correct outdated data.

### Description

- The user should be able to update:
  - first name
  - last name
  - email
  - password
  - profile picture

### Acceptance Criteria (AT-1.2 #130)
Given that the user is logged in as a system administrator and selects the _Settings_ side menu option, 

1. The user should be presented with a form containing their previously supplied user information. 

2. The user selects the _Edit_ button to allow the fields to accept input.

3. Once the information is modified, the user selects the _Save_ button to store the updated data.

### Related Mockups

- System administrator logs into ALTA
![image](https://user-images.githubusercontent.com/19229712/100304623-21351600-2f6d-11eb-9f5f-6fb469b27410.png)

- _Settings_ page - Update account information
![image](https://user-images.githubusercontent.com/19229712/100409284-75a5c780-3039-11eb-92d1-f62048d7eb29.png)

PR #116, opened 2020-10-31
Issue #43 Update profile informations and change password backend

RobertBeaudenon:
Implemented the following features:
- Renamed the registration view to CustomUserView because it is now implementing the functionality of retrieving a single user info
- Created a view to update profile information
- Created a specifc view to change password because it requires custom logic in order to hash new password before saving to db
- Added custom permissions to CustomUserView and to the other views in order to restrict the following:
  - A user can only retrieve his own information (no other user can access the info of others)
  - A user can only update his own info and change his own password
- Tested all of the above

Screen Shot 2020-10-31 at 1 51 56 PM
Issue #43


NicholasNagy:
Most things look fine, I think the only detail I'd like to discuss is how we will be resetting passwords.
Currently, the way I have it set up, is that although in the form, a user can edit another user's password, the changes are not sent to the backend. The reason I did it like this, is because I'm not 100% sure how we should be handling passwords.
I feel like it could make sense to allow SA's to modify passwords, but we can't be careless either, because it could be easy to make a mistake.
Before developing further, we may want to consult with Ajay first before fully implementing how we will manage this API on the frontend.

RobertBeaudenon:
> Most things look fine, I think the only detail I'd like to discuss is how we will be resetting passwords.
> Currently, the way I have it set up, is that although in the form, a user can edit another user's password, the changes are not sent to the backend. The reason I did it like this, is because I'm not 100% sure how we should be handling passwords.
> I feel like it could make sense to allow SA's to modify passwords, but we can't be careless either, because it could be easy to make a mistake.
> Before developing further, we may want to consult with Ajay first before fully implementing how we will manage this API on the frontend.

I understand your concerns, I think for now we should avoid allowing SA modifying the password of other users. So inside the form that I was just checking you could just remove the password field from there and we would have that in a separate independent form just under it in order to have 2 input fields (new password, confirm password) and then we submit it. I have a specific API endpoint in the backend that is just responsible for changing the password ( /changePassword/id/) and this endpoint is currently protected to just allow the owner of the account to change his own password. In the future what we could do, I have a forgot password link in the login page, and once clicked it will generate a temp password that it will send to the user by mail or something around that, but again we will need to confirm with AJAY. So now the only functionality that we will be providing is that he can only change his password if the user is logged in and nobody else could do it for him.

NicholasNagy (changes requested):
I left a comment regarding how we're setting up the permissions.
I recommend splitting up the permissions in a way to make them more reusable.

server/user_account/permissions.py
```diff
@@ -0,0 +1,23 @@
...
+             :param view: Where i can retrieve the arguments passed in url, actions ..
+             :return: True/False : Whether the user is allowed to perform CRUD
+        """
+        if view.action in ['retrieve']:
+            return IsCurrentUserTargetUser.has_permission(None, request, view)
+
+        if view.action in ['create']:
+            return IsSystemAdmin.has_permission(None, request, None)
+
+        return False
```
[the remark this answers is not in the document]
RobertBeaudenon: So I'll explain the reasoning behind that. Restricting to specific actions is only related to ViewSets if you include it inside the permission class.  Whatever permission is reusable is whatever is under django_server/permissions.py where you can find IsSystemAdmin, IsInventoryManager and IsCurrentUserTargetUser. What you see here is a permission that is custom for the registrationViewSet( renamed  CustomUserView). So here It's not generalized to all the view set because some of them might have different permissions depending on their actions. For now i restricted on this view that only the user can retreive his own info but what i can do is allow also the admin to retreive the info of other people (so i can add that). And for registering a user im just restricting it to admins because we didnt start implementing the other epics where IM can also create accounts. So if we need another custom permission for a different viewset we can just create another one based on our needs, because all the viewsets wont have the same permissions based on their different actions. So we could keep it like that. What do you think?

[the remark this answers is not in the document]
NicholasNagy: I see what you mean, but I'm always going to question when someone is making something that isn't reusable.

From what I see though, is we're creating another class to simply provide the proper permissions for another class. This couples the 2 classes together (which are in different files) and reduces cohesion for the class that we using these permissions for. I was wondering if we could do something like this:
viewsetpermissions

This would allow us to customize the permissions for the viewset, without having to create a new class altogether. The code will be much cleaner this way, as we would have the permissions customized in a method within the class as opposed to a method in another class in another file.

Have you tried something like this?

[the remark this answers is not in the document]
RobertBeaudenon: So I've looked into it. It is basically the exact same logic of overriding get_serializer_class() to add different serializers per action which I already made modification for in this commit [2ab2a97](https://github.com/Fruity-Loops/ALTA/pull/116/commits/2ab2a97da2d0420cbd3701c80902f190f54dde0b). So I go proceed and do the same thing with what you mentioned above but I still prefer the original way I did it because it's cleaner I feel. By cleaner, I mean that whatever is related to permissions is just in permissions.py and we can create our custom permissions there without having to write inside the view the code which will end up with a huge viewset. This is my opinion, what do you think.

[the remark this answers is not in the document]
NicholasNagy: I mean, I get the idea of putting all permissions in one place. 
My concern though is understandability. People looking through the permissions file will see this permission and might wonder why it's implemented like this. Most of the other permissions are implemented in a generic fashion for reusability, except for this one. It only has a single use case for a particular view, which can be confusing for someone who is not familiar with what it's being used for.

In the inverse case, someone might be trying to use this view for some functionality and might not have seen/understood the extra custom class. The result might be someone who gets stuck on this view for hours before understanding some silly permissions that they weren't considering.

I understand the concern for a god class, but I really don't think we've arrived there yet.
I know the issues I brought up sound trivial, as someone who knows how to use their IDE can quickly navigate through the files easily, but to me, it just seems like an unnecessary extra step, and decreases the cohesion of the class. 

If you're still opposed to implementing it though, we can ask for a third opinion.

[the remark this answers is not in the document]
RobertBeaudenon: I understand your point, but I still feel it makes the view a bit messier. Also, I implemented the permissions in organization app this way, but still, it can be refactored but I'd be interested what @makchamp thinks since he will be also working intensely on the backend.

client/src/app/components/employee-settings/employee-settings.component.ts
```diff
@@ -70,15 +83,32 @@ export class EmployeeSettingsComponent implements OnInit {
...
-    });
+    this.manageMembersService
+      .updateClientInfo(this.employee, this.id)
+      .subscribe((response) => {
+        location.reload();
+      });
   }
 
+  submitPassword(): void {
+    this.employee.is_active = this.isActive === 'active';
```
NicholasNagy, line 100: why do we need this here, if we're not submitting it?

client/src/app/components/login/login.component.ts
```diff
@@ -75,7 +77,7 @@ export class LoginComponent implements OnInit {
 
   resetForm(): void {
     this.loginForm.reset();
-    Object.keys(this.loginForm.controls).forEach(key => {
+    Object.keys(this.loginForm.controls).forEach((key) => {
       this.loginForm.controls[key].setErrors(null);
     });
```
NicholasNagy, line 82: can we actually make this a single line instead?

For example:
```JavaScript
Object.keys(this.loginForm.controls).forEach(key => this.loginForm.controls[key].setErrors(null));
```

client/src/app/components/employee-settings/employee-settings.component.html
```diff
@@ -13,25 +13,20 @@
...
             <mat-label for="email">Email</mat-label>
             <input matInput id="email" type="text" [(ngModel)]="employee.email" email [disabled]="!edit"/>
           </mat-form-field>
 
-          <mat-form-field>
-            <mat-label for="password">Password</mat-label>
-            <input matInput id="password" type="password" [(ngModel)]="password" minlength="5" [disabled]="!edit"/>
-          </mat-form-field>
-
-          <mat-form-field>
```
NicholasNagy, line 34: there's added functionality for adding a password, so why are we removing it from the form?

NicholasNagy (commented):
In general, it's good, but please see my comments.

client/src/app/components/login/login.component.ts
RobertBeaudenon: I'm not sure i'm gona try, i didn't add that i think it was @makchamp .

client/src/app/components/employee-settings/employee-settings.component.html
RobertBeaudenon: No, because the update password uses a different endpoint. So it needs to be submitted in a different form. Also, later we might add a field another field to confirm the password. I feel its way better this way.

client/src/app/components/employee-settings/employee-settings.component.ts
RobertBeaudenon: I deleted it.

client/src/app/components/login/login.component.ts
RobertBeaudenon: I will keep it this was actually. It is not a big deal for now.

NicholasNagy: it's just a personal preference tbh, it doesn't change the functionality at all

client/src/app/components/employee-settings/employee-settings.component.html
NicholasNagy: oooh, I didn't notice that you added a new form

client/src/app/components/sidenav/sidenavListing.ts
```diff
@@ -1,20 +1,51 @@
...
+    title: 'Manage Organizations',
+    routerLink: 'manage-organizations',
+    subMenuOptions: [],
   },
   {
-    title: 'Manage Members', routerLink: 'manage-members', subMenuOptions: [
-      { title: 'Create Members', routerLink: 'create-members', subMenuOptions: [] },
-      { title: 'Modify Members', routerLink: 'modify-members', subMenuOptions: [] },
-    ]
+    title: 'Edit Profile',
```
NicholasNagy, line 47: Please rename this to 'Settings' like in the mockups

client/src/app/components/employee-settings/employee-settings.component.html
```diff
@@ -13,12 +13,12 @@
...
           <mat-form-field>
             <mat-label for="First Name"></mat-label>
-            <input matInput id="first_name" type="text" [disabled]="true" [value]="employee.first_name"/>
+            <input matInput id="first_name" type="text" [disabled]="!isLoggedInUser"  [value]="employee.first_name"/>
           </mat-form-field>
 
           <mat-form-field>
             <mat-label for="Last Name"></mat-label>
-            <input matInput id="last_name" type="text" [disabled]="true" [value]="employee.last_name"/>
+            <input matInput id="last_name" type="text" [disabled]="!isLoggedInUser" [value]="employee.last_name"/>
```
NicholasNagy, line 21: What's the reasoning behind this change?

RobertBeaudenon: The user must be able to change his first name or last name if he wishes.

NicholasNagy: Didn't we agree to something else in the previous PR?

server/user_account/permissions.py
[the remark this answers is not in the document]
makchamp: Hey guys, for this I agree with both of your perspectives. I would say for me it would largely depend on whether or not you think these permission classes will be reused elsewhere in the future. If they will be used in multiple areas in the app then yes it would probably be more responsible to have them abstracted into their own classes. However, if they're strictly related to an appropriate singular viewset, then I would not bother and implement it the way nick described.

client/src/app/components/employee-settings/employee-settings.component.html
RobertBeaudenon: We agreed for it regarding a SA editing other users and the functionality is the same, but we didn't mention the case of editing its own profile and this is where it is changing.

case1:
Screen Shot 2020-11-01 at 8 47 32 PM

case2:
Screen Shot 2020-11-01 at 8 47 47 PM

NicholasNagy: yeah, I know, I just don't get why it should be different for the user's own profile. BTW, it should still be disabled when the edit mode is off (like in your picture).

TBH, let's just keep this simple. For now, let's just allow everyone to edit the first name and password (which is contradictory to my previous point in the other PR). The API allows it anyway, so it's not like it's secure or anything. 

If it needs to be changed in the future, then we'll just modify it then. For now let's just add it to the question bank that we have for Ajay (along with how we should handle passwords).

NicholasNagy (approved):
Everything looks good now, thanks for the PR, good work!

PR #119, opened 2020-11-01
Registration ui update

Averynder:
This changes the registration page to now match the mockups as close as possible. It should be noted employeeID is the username. None of the core functionality previously developed has changed only the layout. It should also be noted that % were used as much as possible to allow flexibility depending on the size of the screen.


NicholasNagy (changes requested):
Overall, nice job. There's just a few things that need to be fixed.
For starters, you added another top app bar nested within the page. I see what you were trying to do, but we still have the old top app bar there already, and this isn't consistent across the app. However, I think that is an issue for a separate PR to handle, so you can just remove that part.

For the display picture, it would be cool if you could round the image like in the pictures as well (should be easy to do with some css).

Lastly, on higher resolution and bigger screens, the layout becomes inconsistent and ugly. See this picture: 
Screenshot from 2020-11-01 14-05-14

Please use the following [link](https://stackoverflow.com/questions/45350716/detecting-real-time-window-size-changes-in-angular-4) to get the window size, and make a check that will modify the css depending on the window size.

NicholasNagy (approved):
Everything looks good now, thanks for the good work!

Issue #42, opened 2020-10-16
User Story 2.1


Amanda-Kolopa:
As an inventory manager, I want to be able to create a user account for another inventory manager so that I can provide access to initiate audits.

### Description:

- Modify the structure of the code right now to allow inventory managers to log in and create users.
- Add a permission class specific for inventory managers in the backend and allow logging in and creating users.

### Acceptance Criteria ()

1. Log in to an account of an Inventory Manager.
2. Select the Manage Employees tab.
3. At the top right of the page, select the Create button.
4. Fill in fields to create a user and select the Submit button.
5. The new user should be added to the system. To verify, go back to the Manage Employees tab and see if there is an employee with the info that was filled in.

### Related Mockups:

- For logging in to the app
![image](https://user-images.githubusercontent.com/40705372/100802949-48219b00-33f8-11eb-9732-bbde23a216ad.png)

- The Employees tab to manage the employees should look like this one
<img width="630" alt="Screen Shot 2020-12-01 at 5 13 26 PM" src="https://user-images.githubusercontent.com/40705372/100803081-87e88280-33f8-11eb-99f1-5668bad0d948.png">

- The form to fill out to create a new user.
<img width="630" alt="Screen Shot 2020-12-01 at 5 14 24 PM" src="https://user-images.githubusercontent.com/40705372/100803161-ac445f00-33f8-11eb-9b61-6ca4f028cee1.png">

Amanda-Kolopa:
### Progress update:

- set up service to get logged in user's role from the backend
- based on the response, the frontend will remove the option 'System Admin' from the role dropdown menu
- discussed with Robert possible improvements to retrieving data (Angular Directives, global variables)

### Remaining work:

- research mentioned improvements and ask a fellow member to review
- implement tests and add needed comments

### Expected delivery date: 

- latest November 1 (due to midterm season)

PR #125, opened 2020-11-02
Issue #42 - alter front-end view and account creation for IM role

Amanda-Kolopa:
Implemented functionalities:

- in AuthService

  - created BehaviorSubject objects to be used as Observables for the logged-in user's data
  - combined Observers into one (sharedUser) for components to access the logged-in user's data 
  - created a service (getCurrentUser) to retrieve data from CustomUserView
  - removed local storage for logged in user's data (except user_id) 

- in components Toolbar, Sidenav, Signup, Login:

  - setup subscribe to get the user data and alter the front-end view based on the role (SA or non-SA)
  - set onDestroy to unsubscribe (avoid memory leaks)

Notices:

- based on current permission settings and using local storage to hold the logged-in user's id

  - a logged-in SA can alter the local storage id value, refresh the page, and be logged in as a different user (will create a bug report for this)

- the team has discussed improvements to passing the logged-in user's data (consider establishing a parent/child relationship with components to avoid repetitive subscriptions)


NicholasNagy (changes requested):
So there's a few major issues that I can pinpoint right now.

The first is that we are removing the organization option from the create page. Albeit, this is a move in the right direction, but for this to work, many other things must first align. We still need to assign an organization to IM and SK, and now in the way it's made, we're not exactly able to do that as a SA. The idea is that we would click on an organization, get redirected to the organization's page with their dashboard, audits, and templates. In this frame, the SA can make employee accounts for that organization. When they are not in an organization, the SA can only Manage Organizations and Manage other SA. Right now, in the way we have, we can't simply remove organizations, because we won't be able to link an organization's employees to their organization.

Secondly, The user story to be completed was for an IM to create accounts for other IMs. Some logic needs to be made in the backend to allow this. Additionally (this was mine and Avery's mistake), the form is not submitted when you press save, but to fix this, you just need to move the html for the button into the form.

Thirdly, right now, when I login, no matter what I do, the settings page will attempt to display the form for the first user I logged in with. If I log out and log back in, nothing changes. I have no clue what the problem is for this, but we can open a bug report if you don't have time to fix this.

So anyway, in regard to my first point for the organizations, you can either work on them or not depending on if we have time, but regardless, protections need to be placed in the backend to avoid employees making organizations across organizations.

client/src/app/services/auth.service.ts
```diff
@@ -22,4 +59,30 @@ export class AuthService {
...
+      this.role.next(nextRole);
+      this.organizationId.next(nextOrgId);
+      this.organization.next(nextOrg);
+
+      localStorage.setItem('id', nextUserId);
+  }
+
+  setLogin(loggedInUserId: any, loggedInUser: any, loggedInRole: any, loggedInOrgId: any, loggedInOrg: any): void {
+     this.setNext(loggedInUserId, loggedInUser, loggedInRole, loggedInOrgId, loggedInOrg);
+  }
```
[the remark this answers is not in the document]
Amanda-Kolopa: Yes definitely! Not sure what my intentions were for this.
Changes made in commit e4cfcc0

client/src/app/components/signup/signup.component.html
```diff
@@ -103,19 +103,6 @@
...
-            <mat-form-field appearance="outline">
-          <mat-select formControlName="organization">
-            <mat-option value="">Choose an organization</mat-option> <p class="blank-colour">(optional)</p>
-            <mat-option *ngFor="let organization of organizations" [value]="organization">
-              {{ organization.org_name }}
-            </mat-option>
-          </mat-select>
-        </mat-form-field>
-        </div>
-        </mat-grid-tile>
```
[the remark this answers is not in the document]
Amanda-Kolopa: Ok I reverted back to master code

server/user_account/views.py
```diff
@@ -64,21 +67,25 @@ def create(self, request, *args, **kwargs):
         }
 
         auth_user = CustomUser.objects.get(user_name=auth_content['user'])
-        # TODO: Limit account creation by role
-        if auth_user.role != 'SA':
-            pass
+        # Limit account creation by role
+        if (auth_user.role == 'SA') or (auth_user.role == 'IM' and (auth_user.organization_id == request.data.get('organization', ''))):
```
RobertBeaudenon, line 71: Since we are working with permissions here, and the way we have it now is that all the permissions are defined inside get_permissions(self) method, I feel it would be better to keep handling them there:
- So for verifying that the user doing the request is an SA you already have the IsSystemAdmin.
- For verifying that the user is an IM and is trying to add someone in his own organization then you could create a helper for that i think that can be generalized since all the operations that an inventory manager will do will only be allowed in his organization. You can inspire yourself of what i did for IsSystemAdmin and IsInventoryManager.

What do you think @NicholasNagy ?

Amanda-Kolopa: Good catch!
I tried to implement what you described, please let me know if there are possible improvements.

NicholasNagy (changes requested):
Overall, this is looking like it's on the right track. Just a couple of small minor changes and we're good to go. Let me know if you disagree with any of my comments. 

Thanks for the good work!

server/user_account/views.py
NicholasNagy: I agree @RobertBeaudenon , there were just a couple of extra small improvements that I asked @Amanda-Kolopa to do, but aside from that, this PR is looking good.

NicholasNagy (approved):
Everything looks, good!
Thank you! Let's merge it.

Issue #136, opened 2020-11-07
AT-1.6: System administrator modifies an organization's employee's account status


Amanda-Kolopa:
### User Story: ALTA-1.6 (#44)
As a system administrator, I want to be able to modify the status of a user account so that I can update their permissions.

### Acceptance Criteria
Given that the user is logged in as a system administrator and has selected an organization from the _Manage Organizations_ page,

1. The user selects the _Employees_ side menu option to present a table containing the affiliated employee user accounts' information.

4. The user selects the _grey dot_ button at the end of the respective employee's row to present the form containing the employee's account information.

5. The user selects the _Edit_ button to allow the fields to accept input. The user is able to alter the employee account's status to "active" or "disabled". 

6. Once the information is modified, the user selects the _Save_ button to store the altered status. The form is then refreshed to display the updated information.

ajaykaushik17:
Accepted

Issue #135, opened 2020-11-07
AT-1.5: System administrator modifies an organization's employee's account information


Amanda-Kolopa:
### User Story: ALTA-1.5 (#39)
As a system administrator, I want to be able to modify a client’s employee’s user account so that I can reflect any changes requested to their account information.

### Acceptance Criteria
Given that the user is logged in as a system administrator,

1. The user selects an organization from the _Manage Organizations_ page.
2. The user selects the _Employees_ side menu option.
3. The user is presented with a table containing an overview of the affiliated employee user accounts' information.
4. The user selects the _Settings_ button (grey dot) at the end of the respective employee's row.
5. The user is presented with a form containing the employee's account information and selects the _Edit_ button.
6. Once the information is modified, the user selects the _Save_ button to store the updated data. The form is then refreshed to display the updated information.

ajaykaushik17:
need to remove the option to change an employee's role to system admin
also, remove the option to change their password
Consider employee's location aspect for audit assignment 
Status - Pending

ajaykaushik17:
ACCEPTED

Issue #134, opened 2020-11-07
AT-1.4: System administrator create organization's employee's accounts


Amanda-Kolopa:
### User Story: ALTA-1.4 (#33)
As a system administrator, I want to be able to create my client’s employee’s user account so that I can grant their employee access to ALTA.

### Acceptance Criteria
Given that the user is logged in as a system administrator and is viewing the _Manage Organizations_ page,

1. The user is able to select an organization by clicking on its respective row in the presented table. 

2. The user selects the _Employees_ side menu option.

3. The user is presented with a table displaying the affiliated employee user accounts and a _Create_ button. 

4. The user selects the _Create_ button to display an empty form for the user to input the new user’s information.

5. Once the form is complete, the user submits the data by clicking the _Save_ button. The user is then redirected back to the table to validate the creation of the user.

ajaykaushik17:
Accepted - but in the table displaying an organization's employee's, display the user's role
Also Add Employee's location to map audit assignment (Specially for Stock keepers)

Issue #133, opened 2020-11-07
AT-1.3A: System administrator adds an organization


Amanda-Kolopa:
### User Story: ALTA-1.3A (#71)
As a system administrator, I want to be able to add a new organization to ALTA so that I can build my clientele. 

### Acceptance Criteria
Given that the user is logged in as a system administrator,

1. The user should be redirected to the _Manage Organizations_ page which displays a table containing all organizations subscribed to ALTA.
2. The user selects the  _Add_ button to redirect to the _Organization Creation_ page.
3. The user enters the organization's name in the input field.
4. The user submits the new organization by selecting the _Save_ button.

ajaykaushik17:
•	need to be able to include the organization's location upon creation and edit
•	instead of deleting an organization, change status to disabled or inactive

Status - Pending

ajaykaushik17:
ACCEPTED

PR #140, opened 2020-11-08
System (end-to-end) testing setup and conforming acceptance tests for Epic 1

makchamp:
####  - Tool used for e2e testing : JASMINE + Protractor 
Protractor comes packaged with angular. This tool runs selenium engine and allows for complete automation of the chrome browser. In other words it simulates a user’s interaction with an Angular application running in a browser or mobile device.
https://www.protractortest.org/#/infrastructure

#### - Setup environment to run the tests
Since these tests require data to be created, edited and persisted in a database for the duration of the tests, I needed to find a way to prepopulate and flush a test database before and after each test run. There is no direct way for protractor to communicate with django unless done through an API. To avoid extending the api simply for managing these tests, I chose to write a bash script to automate the process.

To run the end-to-end tests (from the root ALTA directory):
**script/e2e**

It will do everything required (no need for any manual configuration)
First it will populate a 'alta_test' database with the values from user fixtures
Then it will run the angular protractor e2e tests
Upon finishing, the test database will be flushed and the django server running on port 8000 will be killed.

#### - Included System tests that conform to the acceptance tests for Epic 1
- AT-1.1: System administrator creates a system administrator account #129
- AT-1.2: System administrator modifies their account information #130 
- AT-1.3: System administrator manages organizations #133 
-  AT-1.4: System administrator create organization's employee's accounts #134 
-  AT-1.5: System administrator modifies an organization's employee's account information #135 
- AT-1.6: System administrator modifies an organization's employee's account status #136

The tests should conform very closely to the acceptance criterias. However they do not deviate much beyond them.
The tests use a Page Object Model. This simply means each page in the app will have its own po file which includes all functions responsible for extracting elements from the DOM. This greatly improves readability, maintainability and reduces code duplication.
Also to note, the order of the tests is very important. I tried as much as possible to decouple the tests (further work might be needed on that front). The order is set in the file _'protractor.conf.js'_

#### - Miscellaneous 
- Created an env variable for the root url of the django api. Extracted all instances where it is used.
- Added user fixtures (users.json) as mentioned above:
This fixture can have many purposes. These users can be used in unit tests as test users (instead of having to type them out manually like we've been doing)
Also they can be inserted in the database, which is useful for system tests (as we need initial users in the database to perform the tests)

       To insert them in the database run the command:
       python manage.py loaddata users.json

        These users were created from the app and then the data was dumped to the file like so:
        python manage.py dumpdata user_account.customuser --indent 4 > users.json

- There is a 1 second delay between the tests. This is mainly done for visualization purposes but I also noticed that it has the added benefit of allowing pages to completely load before transitioning to the next test (prevents errors)
 - Fixed a couple bugs including disabling in the profile settings the first name and last name input fields (since they cant be changed) and allowing the email to be changed.

e2e_preview


Neeham:
Excellent job with the script! The script works perfectly fine however on Windows the source path needs to be changed.

Since this script is based on linux the activate file in venv path is as follow: 

/venv/**bin**/activate

On Windows the activate path in venv is the following:

/venv/**Scripts**/activate

For now, we can replicate 2 copies of e2e file and rename them as follow: e2eLinux and e2eWin specifying the 2 different respective path. In the future we can make a fully working windows script which also closes the open port 8000 upon termination like it does on Linux.

makchamp:
> Excellent job with the script! The script works perfectly fine however on Windows the source path needs to be changed.
> 
> Since this script is based on linux the activate file in venv path is as follow:
> 
> /venv/**bin**/activate
> 
> On Windows the activate path in venv is the following:
> 
> /venv/**Scripts**/activate
> 
> For now, we can replicate 2 copies of e2e file and rename them as follow: e2eLinux and e2eWin specifying the 2 different respective path. In the future we can make a fully working windows script which also closes the open port 8000 upon termination like it does on Linux.

I added another script called 'e2e-win' for windows. Its still bash so its not fully compatible but later we can rewrite it in batch/powershell if we need to. I also changed how the django server is stopped so maybe it will work for you now.

NicholasNagy:
Excellent job! Script works perfectly fine on Linux. For future sprint we can definitely make the script compatible on Windows as well.

NicholasNagy (approved):
Excellent job! Script works perfectly fine on Linux. For future sprint we can definitely make the script compatible on Windows as well.

Issue #45, opened 2020-10-18
User Story 2.2


NicholasNagy:
As an inventory manager, I want to be able to modify my account information so that I can correct outdated data.

### Description:

When logged in as an inventory manager, the user should be able to access some sort of settings page, similar to the pages previously created to allow modification of other employees, that would allow them the same capability for their own information.

### Acceptance Test ():

- Login to an account of an inventory manager
- In the sidenav bar, there is a settings tab, that when you click on it, it will take you to a page with all your information
- The user should then be able to modify the fields in the form of information by clicking on the edit button at the top right
- Then, the user should be able to save their changes to be submitted, and the page will be refreshed to show the updated info

### Related Mockups:

- The page to login to the app
![image](https://user-images.githubusercontent.com/40705372/101215517-fbce9900-364b-11eb-9649-d79d74832bb5.png)

- The page upon logging into the app. On the side navigation, there should be a tab for user settings
![image](https://user-images.githubusercontent.com/40705372/101215598-1dc81b80-364c-11eb-8fe2-61ba894f5113.png)

- On this page, the settings tab is highlighted and the profile of the user is shown with a button to begin editing
![image](https://user-images.githubusercontent.com/40705372/101216784-1ace2a80-364e-11eb-99b4-6523288a2a25.png)

- The associated fields are now editable and can be modified for the user to change their information
![image](https://user-images.githubusercontent.com/40705372/101267488-46711380-3727-11eb-9ba6-f957e6cfe4a4.png)

PR #141, opened 2020-11-10
IM can modify their settings

Averynder:
This adds a simple check and if the user is an IM, they can access the previously locked fields to modify their settings. This is for User story 2.2 / issue #45 .

### DIFF
--- a/client/src/app/components/employee-settings/employee-settings.component.html
+++ b/client/src/app/components/employee-settings/employee-settings.component.html
@@ -14,7 +14,7 @@
           <div class="form-group">
             <mat-label for="power">First Name:</mat-label>
             <mat-form-field appearance="outline">
-              <input matInput id="firstname" type="text" [disabled]=true [value]="employee.first_name"/>
+              <input matInput id="firstname" type="text" [disabled]="!canEditFields || !edit" [value]="employee.first_name"/>
             </mat-form-field>
           </div>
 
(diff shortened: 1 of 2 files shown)


client/src/app/components/employee-settings/employee-settings.component.html
```diff
@@ -14,7 +14,7 @@
           <div class="form-group">
             <mat-label for="power">First Name:</mat-label>
             <mat-form-field appearance="outline">
-              <input matInput id="firstname" type="text" [disabled]=true [value]="employee.first_name"/>
+              <input matInput id="firstname" type="text" [disabled]="!canEditFields || !edit" [(ngModel)]="employee.first_name"/>
```
NicholasNagy, line 17: why?

client/src/app/components/employee-settings/employee-settings.component.html
```diff
@@ -33,7 +33,7 @@
             <mat-label for="lastname">Last name:</mat-label>
             <mat-form-field appearance="outline">
               <input matInput id="lastname" type="text" placeholder="Last name"
-                     [disabled]=true [value]="employee.last_name"/>
+                     [disabled]="!canEditFields || !edit" [(ngModel)]="employee.last_name"/>
```
NicholasNagy, line 36: why?

NicholasNagy (changes requested):
I don't think we should be enabling these fields at all, according to our discussions.

NicholasNagy:
As per our discussion last night, this PR will be closed as its features aren't intended

PR #155, opened 2020-11-18
Im settings permissions

Averynder:
This adds the fixed permissions based on the organization submitted and the way it is sent matching the appropriate function in the backend with the request. It also adjusts the tests accordingly

### DIFF
--- a/server/user_account/views.py
+++ b/server/user_account/views.py
@@ -7,7 +7,8 @@ from rest_framework import status, viewsets, generics
 from rest_framework.authtoken.models import Token
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
-from django_server.permissions import IsSystemAdmin, IsCurrentUserTargetUser, IsInventoryManager
+from django_server.permissions import IsSystemAdmin, IsCurrentUserTargetUser, IsInventoryManager,\
+    get_self_org, get_self_org_query, get_self_org_body
 from .serializers import UserSerializer, LoginSerializer, ClientGridSerializer,\
     UserPasswordSerializer
 from .models import CustomUser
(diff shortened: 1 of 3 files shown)


server/user_account/views.py
```diff
@@ -44,13 +45,24 @@ def get_permissions(self):
...
+        except CustomUser.DoesNotExist:
+            permission_classes = [IsAuthenticated, IsSystemAdmin]
 
-        # TODO: Validate requested user id matches requested organization in DB
-        # for permissions unrelated to create
-        elif self.action in ['create', 'list']:
+        if self.action in ['create']:
+            permission_classes = [IsAuthenticated, IsInventoryManager | IsSystemAdmin]
+        elif self.action in ['list']:
+            get_self_org_query(user, self.request)
```
NicholasNagy, line 57: this line does nothing

server/user_account/views.py
```diff
@@ -44,13 +45,24 @@ def get_permissions(self):
...
             permission_classes = [IsAuthenticated, IsInventoryManager | IsSystemAdmin]
+        elif self.action in ['retrieve', 'update']:
+            permission_classes = [IsAuthenticated, IsCurrentUserTargetUser, IsInventoryManager\
+                                  | IsSystemAdmin]
+        elif self.action in ['partial_update']:
+            get_self_org_body(user, self.request)
```
NicholasNagy, line 63: this line doesn't do anything

server/user_account/views.py
```diff
@@ -44,13 +45,24 @@ def get_permissions(self):
...
-        # for permissions unrelated to create
-        elif self.action in ['create', 'list']:
+        if self.action in ['create']:
+            permission_classes = [IsAuthenticated, IsInventoryManager | IsSystemAdmin]
+        elif self.action in ['list']:
+            get_self_org_query(user, self.request)
             permission_classes = [IsAuthenticated, IsInventoryManager | IsSystemAdmin]
+        elif self.action in ['retrieve', 'update']:
+            permission_classes = [IsAuthenticated, IsCurrentUserTargetUser, IsInventoryManager\
+                                  | IsSystemAdmin]
```
NicholasNagy, line 61: you would also need to check the org here

server/user_account/views.py
```diff
@@ -44,13 +45,24 @@ def get_permissions(self):
...
+        try:
+            user = CustomUser.objects.get(email=self.request.user)
+        except CustomUser.DoesNotExist:
+            permission_classes = [IsAuthenticated, IsSystemAdmin]
 
-        # TODO: Validate requested user id matches requested organization in DB
-        # for permissions unrelated to create
-        elif self.action in ['create', 'list']:
+        if self.action in ['create']:
+            permission_classes = [IsAuthenticated, IsInventoryManager | IsSystemAdmin]
```
NicholasNagy, line 55: You would also need to check the org here

NicholasNagy (changes requested):
To sum up, your changes loosen the permissions in every single aspect. There isn't a single instance where you are validly checking that the organization matches the user.
Also, you forgot to check the organizations views. These changes also affect those.
We will discuss further tonight.

NicholasNagy:
Closing this PR in favour of #159 for its better overall implementation of the fix for this issue.

Issue #46, opened 2020-10-18
User Story 2.3


NicholasNagy:
As an inventory manager, I want to be able to access the inventory data of the company’s ERP system through ALTA so that I can view potential items to audit. 

### Description:

This user story will include loading up a csv file from a shared folder and inserting the contents (which should be the inventory extract) into the db. Since we're still going to be hashing out the details of how we can access this shared folder, for now we can just hard-code the dummy file and load that up by default. It's important to design this feature with extensibility in mind, as we will need to add more appropriate functionality later on depending on the method we decide to go with.

Additionally, this user story will need to display the results on the front end in a tabular format.

### Acceptance test ():

- Log in as an inventory manager
- In the sidenav, there should be an inventory items tab to click on.
- Once clicked, there should be a table displaying all the inventory items collected from the data dump for the organization.

### Related Mockups:

- The user must log in as an inventory manager on this page
![image](https://user-images.githubusercontent.com/40705372/101231437-474b6c00-3679-11eb-8b6f-1cefbad89db8.png)

- On the side navigation bar there is a tab called Inventory to click on
![image](https://user-images.githubusercontent.com/40705372/101231423-2c78f780-3679-11eb-870d-719b0c8b1ae4.png)

- The table should look like this, except it should have all the fields from the inventory extract (which can possibly change)
![image](https://user-images.githubusercontent.com/40705372/101231467-7d88eb80-3679-11eb-8591-78fea1445947.png)

PR #156, opened 2020-11-18
Issue #46 Populate inventory data

RobertBeaudenon:
Issue #46 
In this pull request, I wrote a python script that I placed under django_server folder that is triggered one time once the server runs. The file is triggered from the file  django_server/urls.py because this file is initialized once when the server starts.
Along with the file in the same directory, i inserted the CSV containing the dummyData.
This pyscript will be responsible of:
- Creating a connection to the db using pymongo
- Creating a collection that will hold the data populated from the csv which is named with what django refers to in the format of "appname_modelname" so its "inventory_item_item" (This is what django would have called it if we generated it using "python manage.py migrate".
So in order to later create views that will interact with it we must respect this convention.

On top of that I created an app called inventory_item responsible of holding the operations that we do on the items:
- I created an API endpoint responsible for retrieving the items
- For now i'm returning all the fields but this can be modified upon request because it is not yet clear what we want to return to the user


NicholasNagy (commented):
Everything looks good, but weren't we planning to do have the model be dynamic depending on what the collection contains in the DB?

RobertBeaudenon:
> Everything looks good, but weren't we planning to do have the model be dynamic depending on what the collection contains in the DB?

PR #157, opened 2020-11-18
Issue#46 Populate inventory data

RobertBeaudenon:
Issue #46 
In this pull request, I wrote a python script that I placed under django_server folder that is triggered one time once the server runs. The file is triggered from the file  django_server/urls.py because this file is initialized once when the server starts.
Along with the file in the same directory, i inserted the CSV containing the dummyData.
This pyscript will be responsible of:
- Creating a connection to the db using pymongo
- Creating a collection that will hold the data populated from the csv which is named with what django refers to in the format of "appname_modelname" so its "inventory_item_item" (This is what django would have called it if we generated it using "python manage.py migrate".
So in order to later create views that will interact with it we must respect this convention.

On top of that I created an app called inventory_item responsible of holding the operations that we do on the items:
- I created an API endpoint responsible for retrieving the items
- For now i'm returning all the fields but this can be modified upon request because it is not yet clear what we want to return to the user


client/src/app/models/InventoryItem.ts
```diff
@@ -0,0 +1,17 @@
...
+  Part_Description: string;
+  Serial_Number: string;
+  Condition: string;
+  Category: string;
+  Owner: string;
+  Criticality: string;
+  Average_Cost: string;
+  Quantity: string;
+  Unit_of_Measure: string;
+}
```
NicholasNagy, line 17: Again, I think this is something that we need to generate dynamically

NicholasNagy (changes requested):
Overall, I think the look and feel is good.

My one concern is that we are hard-coding the fields of the inventory items. They probably will change, and we will have to modify them, so to avoid mistakes, we should just generate all the columns dynamically. It will save us time and changes in the front end

client/src/app/components/manage-inventory-items/manage-inventory-items.component.ts
```diff
@@ -0,0 +1,63 @@
...
+    'Part_Description',
+    'Serial_Number',
+    'Condition',
+    'Category',
+    'Owner',
+    'Criticality',
+    'Average_Cost',
+    'Quantity',
+    'Unit_of_Measure',
+  ];
```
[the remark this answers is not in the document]
RobertBeaudenon: Definitely I agree. This has been addressed.

client/src/app/models/InventoryItem.ts
RobertBeaudenon: I don't think we need to generate an interface anymore for it in the new implementation.

NicholasNagy (approved):
Everything looks good now, thanks for the changes.

I'll merge now.

PR #159, opened 2020-11-21
Organization Checking

Averynder:
Previously within the Inventory manager we had 3 possible checks for the organization, however they were not combined. This now moves these checks into their own functions and then proceeds to check them based on whatever was sent, regardless of the type of request sent. If 2 different organizations are sent or none at all, then the permission will be denied

### DIFF
--- a/server/django_server/permissions.py
+++ b/server/django_server/permissions.py
@@ -17,9 +17,18 @@ class IsSystemAdmin(BasePermission):
         user = CustomUser.objects.get(email=request.user)
         return user.role == 'SA'
 
+
 def get_self_org(user, request):
     return str(user.organization_id) == request.parser_context['kwargs']['pk'] and request.path.\
         startswith('/organization')
+
+
+def get_self_org_query(user, request):
+    return str(user.organization_id) == request.GET.get("organization", '')
+
+
+def get_self_org_body(user, request):
+    return str(user.organization_id) == request.data.get('organization', '')
 
 
 class IsInventoryManager(BasePermission):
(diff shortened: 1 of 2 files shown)


server/django_server/permissions.py
```diff
@@ -35,16 +44,29 @@ def has_permission(self, request, view):
...
-                get("organization", '')
+        if not sent_proper_organization:
+            return False
 
-        if user.role == 'IM':
-            if request.data.get('role', '') != 'SA' and \
-                    (str(user.organization_id) == request.data.get('organization', '') or
-                     get_self_org(user, request)):
-                return True
+        if user.role in ['IM', 'SA']:
```
NicholasNagy, line 68: we're checking if the user is an inventory manager, so the 'SA' check here is actually incorrect.

server/django_server/permissions.py
```diff
@@ -35,16 +44,29 @@ def has_permission(self, request, view):
...
+        sent_proper_organization = False
+        for found_true in correct_organization:
+            if found_true:
+                sent_proper_organization = True
 
-        if view.action in ['list']:
-            return user.role == 'IM' and str(user.organization_id) == request.GET.\
-                get("organization", '')
+        if not sent_proper_organization:
+            return False
```
NicholasNagy, line 66: My first thought was: what's this for?
But this is actually smart. We should include a comment at the top to explain why we have this.
Maybe something along the lines of: ensuring that only the requests accounted for are being made.

NicholasNagy (changes requested):
Overall, looks good, but I think there is a few details to change.
Please see my comments.
This implementation is much better than #155 so I will close that PR in favor of this one.

NicholasNagy (approved):
Looks good now, once all the tests are passed, I will merge it.

Issue #47, opened 2020-10-18
User Story 2.4


NicholasNagy:
### User Story: ALTA-2.4

As an inventory manager, I want to be able to schedule automatic inventory data updates at specified intervals so that I can view the latest version of the company’s ERP system’s data.

### Description:

An inventory manager is able to set the interval of time for which the system will load up the inventory extract. On the sidebar, the inventory manager can access the Organization Settings page. This page will allow setting the FTP file location, the recurrence of the data updates, and also to upload inventory extract. 

### Acceptance Criteria (AT-2.4 #216):
Given that the user is logged in as an inventory manager,

1. The user should be redirected to the _Dashboard_ page.
2. The user selects the _Organization Settings_ side menu option.
3. The user is presented with a form containing the FTP file location and the extract repetition rate.
4. The user selects the _Edit_ button.
5. The user inputs the designated location of the inventory data and sets the extract rate.
6. The user selects the _Save_ button. The form is then refreshed to display the updated information.

### Related Mockups:

- Inventory manager login page
<img width="659" alt="log in" src="https://user-images.githubusercontent.com/35980232/103579566-c9c58680-4ed8-11eb-946e-f837b6382b7d.png">

- Dashboard page
![dashboard](https://user-images.githubusercontent.com/35980232/112053199-6f78b100-8b2a-11eb-8342-c3160b4b099b.png)

![Capture d’écran 2021-01-14 à 21 06 37](https://user-images.githubusercontent.com/35980232/104671748-6b549100-56ac-11eb-8370-6972f9fbdcdc.png)

### **REMAINING WORK** : 

- Move scheduling inventory updates functionality from the _Inventory Items_ page to the _Organization Settings_ page

PR #162, opened 2020-11-25
Schedule data updates

RobertBeaudenon:
Issue #47 
Implemented a method that updates the inventory items collection at predefined intervals of time. As of now, it is updating the collection every 1 min just for testing purposes.

When you run the backend just execute this command:
- python manage.py runserver --noreload -> the no reload flag will prevent Django from restarting the server each time you save because this is the behavior when DEBUG mode is true, so we do that to mimic a production environment.

To log the operations it will be printed in the console:
- When the collection is first initialized: "Data inserted successfully"
- When the collection is updated: "Data Refreshed" every 1 min


server/inventory_item/updater.py
```diff
@@ -0,0 +1,8 @@
+from apscheduler.schedulers.background import BackgroundScheduler
+from django_server.load_csv_to_db import main
+
+
+def start():
+    scheduler = BackgroundScheduler()
+    scheduler.add_job(main, 'interval', minutes=1)
```
NicholasNagy, line 7: Is this starting a thread or a process? 
Can you explain how a job is implemented in this library?

server/django_server/load_csv_to_db.py
```diff
@@ -41,3 +42,24 @@ def populate_items(csv_file, collection_name):
             # If the error is different to duplicate key entry
             if error.details["writeErrors"][i]["code"] != 11000:
                 print("An error occured")
+
+
+def update_items(csv_file, collection_name):
+    collection = get_collection(collection_name)
+    dict_of_records = clean_data(csv_file)
```
NicholasNagy, line 49: I'm raising this as a topic of discussion, as opposed to a suggestion for this PR (as this is outside the scope of the PR).
I think that we should implement the clean_data method as a generator (lazy loading) to avoid keeping all the data in memory all the time. I'm not sure if pandas does this out of the box, but it seems to me that when you convert it to a dictionary, we would be essentially converting the whole thing into a data structure that is definitely stored in memory.

What are your thoughts? This might be a good dev story for next sprint.

NicholasNagy (changes requested):
Hey, so I left some comments, but there are still 2 main problems that I see with this code right now.

Problem 1: I'm starting to see the direction where we're going with this code right now, and the current design does not easily allow for updating inventory items from multiple organizations. We could separate this out into another dev story for a separate sprint, since Dominik is also probably going to be relying on this design as well. 

What are your thoughts? where should we go from here for this problem?

Problem 2: The current implementation applies no variability to the time interval based upon which each organization may want their inventory to be updated. This is assuming that the interval of the job cannot be modified while it is running, so have you looked into that? 
Either way, we would still need to have the interval times stored somewhere for each organization and loaded up on each job start up.

server/inventory_item/updater.py
RobertBeaudenon: APScheduler has four kinds of components:

- triggers

- job stores

- executors

- schedulers

Triggers contain the scheduling logic. Each job has its own trigger which determines when the job should be run next. Beyond their initial configuration, triggers are completely stateless.

Job stores house the scheduled jobs. The default job store simply keeps the jobs in memory, but others store them in various kinds of databases. A job’s data is serialized when it is saved to a persistent job store, and deserialized when it’s loaded back from it. Job stores (other than the default one) don’t keep the job data in memory, but act as middlemen for saving, loading, updating and searching jobs in the backend. Job stores must never be shared between schedulers.

Executors are what handle the running of the jobs. They do this typically by submitting the designated callable in a job to a thread or process pool. When the job is done, the executor notifies the scheduler which then emits an appropriate event.

Schedulers are what bind the rest together. You typically have only one scheduler running in your application. The application developer doesn’t normally deal with the job stores, executors or triggers directly. Instead, the scheduler provides the proper interface to handle all those. Configuring the job stores and executors is done through the scheduler, as is adding, modifying and removing jobs.

server/django_server/load_csv_to_db.py
RobertBeaudenon: I see what you mean, I have absolutely no clue, but i can start looking into it and let you know the updates. What do you suggest for this user story, should we keep it like that for now?

server/inventory_item/updater.py
NicholasNagy: I get that, but what about the job itself? How is it implemented? What is the underlying python process?

When you start a job, what is the library doing? We need to know this before using the library, because if we have cpu-bound tasks, we can't be relying on this if it's using threads.

server/django_server/load_csv_to_db.py
NicholasNagy: For this user story, it's fine, this is more of a production issue I can see coming.

server/inventory_item/updater.py
RobertBeaudenon: This can be configured. "The default ThreadPoolExecutor should be good enough for most purposes. If your workload involves CPU intensive operations, you should consider using ProcessPoolExecutor instead to make use of multiple CPU cores. You could even use both at once, adding the process pool executor as a secondary executor."

NicholasNagy: okay, awesome, this is what I wanted to know.

RobertBeaudenon: > Problem 1: I'm starting to see the direction where we're going with this code right now, and the current design does not easily allow for updating inventory items from multiple organizations. We could separate this out into another dev story for a separate sprint, since Dominik is also probably going to be relying on this design as well.
> 
> What are your thoughts? where should we go from here for this problem?


Problem 1:
Okay, I see what you mean, but one thing I don't really understand is that the database won't be a centralized one for all the organizations. Can you clarify a bit? Because as of now we just have one inventory items collection, how are we suppose to handle that? Maybe we could discuss it on a call.

RobertBeaudenon:
> Problem 1: I'm starting to see the direction where we're going with this code right now, and the current design does not easily allow for updating inventory items from multiple organizations. We could separate this out into another dev story for a separate sprint, since Dominik is also probably going to be relying on this design as well.
> 
> What are your thoughts? where should we go from here for this problem?



Okay, I see what you mean, but one thing I don't really understand is that the database won't be a centralized one for all the organizations. Can you clarify a bit? Because as of now we just have one inventory items collection, how are we suppose to handle that? Maybe we could discuss it on a call.

NicholasNagy:
> Okay, I see what you mean, but one thing I don't really understand is that the database won't be a centralized one for all the organizations. Can you clarify a bit? Because as of now we just have one inventory items collection, how are we suppose to handle that? Maybe we could discuss it on a call.

We would most likely have a centralized DB for all the organizations. This is why we are able to set up multiple organizations on a single db. All we need for the inventory items is to have a foreign key linking to the id of each organization. If it's not clear, message me on discord and we can maybe get into a call.

NicholasNagy (changes requested):
I left a few design concerns related to the code written. Please have a look and consider my suggestions.

Another thing I want to mention, is that I can't quite wrap my head around how we are starting the jobs. I made a review comment, thinking that it might by an endpoint you created, but can you explain to me what's going?

server/django_server/load_csv_to_db.py
```diff
@@ -41,3 +46,61 @@ def populate_items(csv_file, collection_name):
...
+#         client.force_authenticate(user=system_admin)
+#
+#     # Create an organization that could be linked to the inventory items
+#     try:
+#         organization = Organization.objects.get(org_name='test_organization')
+#         return organization.org_id
+#     except Organization.DoesNotExist:
+#         data = {'org_name': 'test_organization'}
+#         response = client.post("/organization/", data)
+#         return response.data["org_id"]
```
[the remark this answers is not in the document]
RobertBeaudenon: I'll delete it

server/django_server/load_csv_to_db.py
```diff
@@ -25,12 +27,15 @@ def clean_data(csv_file):
...
+    dataframe["organization"] = dummy_org_id
+
     return dataframe.to_dict('records')
 
 
-def populate_items(csv_file, collection_name):
+def populate_items(csv_file, collection_name, dummy_org_id):
     collection = get_collection(collection_name)
-    dict_of_records = clean_data(csv_file)
+    dict_of_records = clean_data(csv_file, dummy_org_id)
```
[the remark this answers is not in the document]
RobertBeaudenon: Because it was related to the function that i will delete, i'm gona update the name

server/organization/models.py
```diff
@@ -1,7 +1,17 @@
...
     org_id = models.AutoField(primary_key=True)
     org_name = models.CharField(max_length=256, unique=True)
     status = models.BooleanField(default=False)
+    inventory_items_refresh_job = models.IntegerField(default=1)
+
+
+@receiver(post_save, sender=Organization)
+def execute_job(instance, **kwargs):
+    print('post save callback')
+    start_new_job(instance.org_name, instance.inventory_items_refresh_job)
```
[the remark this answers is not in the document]
RobertBeaudenon: Yes this is basically how signals works, this method will be triggered once the Organization instance has been saved

RobertBeaudenon:
> Another thing I want to mention, is that I can't quite wrap my head around how we are starting the jobs. I made a review comment, thinking that it might by an endpoint you created, but can you explain to me what's going?

Once the app starts the background scheduler is starting, and then each time an Organization instance is created a job is created along that and added to the scheduler which will take care of executing the jobs depending on their settings.

server/inventory_item/updater.py
```diff
@@ -0,0 +1,66 @@
...
+
+job_stores = {
+    'default': MongoDBJobStore(database="alta_development",
+                               collection="jobs", client=client)
+}
+
+# This will get you a BackgroundScheduler with a MongoDBJobStore
+# named “default” to replace the default one  and a ThreadPoolExecutor
+# named “default” with a default maximum thread count of 10.
+scheduler = BackgroundScheduler(jobstores=job_stores)
```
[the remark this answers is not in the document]
RobertBeaudenon: Yeah, I don't really know how to do that, I suggest to get on a call to explain more your approach. Also I did it this way because i need to access the scheduler globally everywhere

NicholasNagy:
> > Another thing I want to mention, is that I can't quite wrap my head around how we are starting the jobs. I made a review comment, thinking that it might by an endpoint you created, but can you explain to me what's going?
> 
> Once the app starts the background scheduler is starting, and then each time an Organization instance is created a job is created along that and added to the scheduler which will take care of executing the jobs depending on their settings.

Yeah, but what happens when you turn off the application and open it again?

RobertBeaudenon:
> > > Another thing I want to mention, is that I can't quite wrap my head around how we are starting the jobs. I made a review comment, thinking that it might by an endpoint you created, but can you explain to me what's going?
> > 
> > 
> > Once the app starts the background scheduler is starting, and then each time an Organization instance is created a job is created along that and added to the scheduler which will take care of executing the jobs depending on their settings.
> 
> Yeah, but what happens when you turn off the application and open it again?

What do you mean turn off the app? Do you mean stoping the server?  if this is what you meant then all the jobs are stored in the collection so when you start again the server all the jobs will continue running, they will not be lost.

NicholasNagy:
> > > > Another thing I want to mention, is that I can't quite wrap my head around how we are starting the jobs. I made a review comment, thinking that it might by an endpoint you created, but can you explain to me what's going?
> > > 
> > > 
> > > Once the app starts the background scheduler is starting, and then each time an Organization instance is created a job is created along that and added to the scheduler which will take care of executing the jobs depending on their settings.
> > 
> > 
> > Yeah, but what happens when you turn off the application and open it again?
> 
> What do you mean turn off the app? Do you mean stoping the server? if this is what you meant then all the jobs are stored in the collection so when you start again the server all the jobs will continue running, they will not be lost.

That's exactly what I meant, and that answers my question, thank you!

server/inventory_item/apps.py
```diff
@@ -3,3 +3,11 @@
...
 class ItemsConfig(AppConfig):
     name = 'inventory_item'
+
+    # To let Django know it needs to kick off our updater on startup
+    # overwrite the AppConfig.ready() method.
+    def ready(self):
+        #  Due to the intricacies of inheritance, any imports for this
+        #  override must be located within be body of the ready() method
+        from . import updater
+        updater.start()
```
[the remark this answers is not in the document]
NicholasNagy: @RobertBeaudenon , you never addressed this comment about how you are importing the updater

client/src/app/components/toolbar/toolbar.component.ts
```diff
@@ -28,15 +28,17 @@ export class ToolbarComponent implements OnInit {
...
+      this.loggedInUserRole = data.role;
+      this.organization = data.org;
+    });
+    this.orgModeSub = this.authService.getOrgMode().subscribe((value) => {
       this.orgMode = value;
     });
+    console.log(this.orgModeSub);
+    console.log(this.orgMode);
+    console.log(this.organization);
   }
```
NicholasNagy, line 42: why did you add these print statements?

server/inventory_item/updater.py
```diff
@@ -0,0 +1,88 @@
...
+                                       collection="jobs", client=client)
+        }
+
+        # This will get you a BackgroundScheduler with a MongoDBJobStore
+        # named “default” to replace the default one  and a ThreadPoolExecutor
+        # named “default” with a default maximum thread count of 10.
+        self.scheduler = BackgroundScheduler(jobstores=job_stores)
+
+
+scheduler = Scheduler("mongodb://localhost:27017/", "alta_development").scheduler
```
NicholasNagy, line 40: why aren't you including this in the singleton?

server/inventory_item/updater.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def get_specific_job(job_id):
+    job = scheduler.get_job(job_id)
+    print("#############")
+    print("###" + str(job_id) + "### " + str(job))
+    print("#############")
+
+
+scheduler.add_listener(scheduler_listener, EVENT_JOB_EXECUTED | EVENT_JOB_ERROR)
```
NicholasNagy, line 88: Why aren't you including this in the singleton?

server/organization/views.py
```diff
@@ -14,3 +16,28 @@ class OrganizationViewSet(viewsets.ModelViewSet):
...
     serializer_class = OrganizationSerializer
     permission_classes = [IsAuthenticated, UserOrganizationPermission]
+
+
+class ModifyOrganizationInventoryItemsDataUpdate(generics.GenericAPIView):
+    """
+    API endpoint that allow a user to update the timing at which
+    the Inventory Data is refreshed
+    """
+    permission_classes = [IsAuthenticated]
```
NicholasNagy, line 26: you should update the permission to be more restrictive

server/organization/views.py
```diff
@@ -14,3 +16,28 @@ class OrganizationViewSet(viewsets.ModelViewSet):
...
+
+    def post(self, request):
+        data = request.data
+        print(data)
```
NicholasNagy, line 30: why are you logging here, is this intentional?

NicholasNagy (changes requested):
I left a few comments, but there was one thing that I was unsure about.

I didn't really understand what the factory package you're using is for and I was curious to know what it is being used for.

client/src/app/components/toolbar/toolbar.component.ts
RobertBeaudenon: I will delete them.

server/inventory_item/updater.py
RobertBeaudenon: I will do it.

server/organization/views.py
RobertBeaudenon: Okay I wanted to discuss that. So the restrictions should be SA and IM that are just part of the organization right?

server/organization/views.py
RobertBeaudenon: No i forgot to remove it, I'll delete it

RobertBeaudenon:
> I left a few comments, but there was one thing that I was unsure about.
> 
> I didn't really understand what the factory package you're using is for and I was curious to know what it is being used for.

The factory package allows me to disable the signals triggered in the django app in certain test case. So the reason behind disabling them is that when an organization is created we have an automatic job that is specific to that organization that is created.  Since we are creating dummy organizations in the test cases, it will create a bunch of random jobs in our db that we dont really need. So this is why i'm doing that.

server/inventory_item/updater.py
RobertBeaudenon: I can do it, it's just that if we need to update the db name or db url i thought it would just be easier to pass them as arguments instead of going manually change them in the singleton. What do you think?

NicholasNagy: I don't see how having them here changes how you define it at all.

Just define it as a variable in the above class. Or am I missing something?

server/organization/views.py
NicholasNagy: Yep

NicholasNagy:
> > @RobertBeaudenon , you never addressed this comment about how you are importing the updater
> 
> regarding that it is explained in the comment above the import, stating that any import to the override must be inside the method.

I mean, I read that, but I don't see how inheritance is causing this issue. Can you link some documentation?

server/inventory_item/updater.py
RobertBeaudenon: Okay

RobertBeaudenon:
> > > @RobertBeaudenon , you never addressed this comment about how you are importing the updater
> > 
> > 
> > regarding that it is explained in the comment above the import, stating that any import to the override must be inside the method.
> 
> I mean, I read that, but I don't see how inheritance is causing this issue. Can you link some documentation?

It was in this [article](https://medium.com/@kevin.michael.horan/scheduling-tasks-in-django-with-the-advanced-python-scheduler-663f17e868e6)

NicholasNagy:
> It was in this [article](https://medium.com/@kevin.michael.horan/scheduling-tasks-in-django-with-the-advanced-python-scheduler-663f17e868e6)

okay, can you link this in the code, for anyone who is curious?

Also there is now a merge conflict in the tests.

NicholasNagy (approved):
Alright, thank you for fixing up the PR, it's ready to be merged now!

PR #163, opened 2020-11-26
Add base mobile client (@ionic/angular v5.5.1) using Capacitor native runtime 

makchamp:
This is the initial configuration of the mobile client that will be mainly used by stock keepers.
I added native builds for Android and iOS. There is no actual consideration for the native platforms in the code yet so the app will load but logging in will not work (cookies dont work, etc)

I am using Capictor (ionic's official native app runtime) over Cordova. A nice overview of the differences between the two can be read here:
https://ionicframework.com/resources/articles/capacitor-vs-cordova-modern-hybrid-app-development

#### Install ionic tooling: ####  
```npm install -g @ionic/cli native-run cordova-res```
#### Deployment to native platforms: #### 
- First build Web:
    - `ionic build --project="alta-front"`
- Build Android:
    - `ionic cap add android --project="alta-front"` 
Then open 'android' project in Android Studio
- Build iOS:
    - `ionic cap add ios --project="alta-front"`
Then open 'ios' project in XCode (requires mac os)

#### Reload changes: #### 
- Copy changes to native projects:
 `ionic cap copy --project="alta-front"`

Note: Because this project wasn't built with ionic out of the gate, we have to specify the project argument in all the commands. I have tried to find a way (if possible) to tell ionic to look for that project name by default but have not found a solution yet. 
I have tested the Android version. The iOS one should work as well but I have not tested it yet.

source: https://ionicframework.com/docs/angular/your-first-app/6-deploying-mobile


NicholasNagy:
Hey, so everything works well, I'm just curious as to why we are pinging the same login API as the inventory manager and system admin?

What's your plan for logging in?

NicholasNagy (changes requested):
I just want to add a couple of things to your description. Before any of this works, you need to run 

```npm install -g ionic```

so that we can use ionic from the command line.

Another issue I ran into, is that it wasn't building with XCode on my mac, so I had to install cocoa pods.
I followed the following link to be able to run it: [Link](https://stackoverflow.com/questions/57262276/build-failed-unable-to-open-file-in-target-in-project-in-target)

I then deleted the ios directory and rebuilt the app, and it ended up working.

Another point I want to add, is that I don't know what was touched in the web app, but when I run it on this branch and I try reloading the page anywhere within the app, the whole app goes underneath the sidenav bar as if it wasn't there. Can you look into this? I don't have the problem on master.

makchamp:
> Hey, so everything works well, I'm just curious as to why we are pinging the same login API as the inventory manager and system admin?
> 
> What's your plan for logging in?

Yea logging in wont work right now because we cant store tokens inside cookies on mobile obviously
So new login logic has to be written for that especially
I'll do that on another pull request.
Basically the user will be directed to entirely new pages after logging in from mobile devices. The only common page between the web mobile will be the greeting/login page for now

makchamp:
> I just want to add a couple of things to your description. Before any of this works, you need to run
> 
> `npm install -g ionic`
> 
> so that we can use ionic from the command line.
> 
> Another issue I ran into, is that it wasn't building with XCode on my mac, so I had to install cocoa pods.
> I followed the following link to be able to run it: [Link](https://stackoverflow.com/questions/57262276/build-failed-unable-to-open-file-in-target-in-project-in-target)
> 
> I then deleted the ios directory and rebuilt the app, and it ended up working.
> 
> Another point I want to add, is that I don't know what was touched in the web app, but when I run it on this branch and I try reloading the page anywhere within the app, the whole app goes underneath the sidenav bar as if it wasn't there. Can you look into this? I don't have the problem on master.

Thanks for the feedback. I will test it on iOS as well. 
I fixed the issue with the content appearing under the side nav menu 56f60b6fa1dfb47ab4827754f6368ce6f1c6c281

NicholasNagy (approved):
I looked over the changes, the sidenav no longer has the issue, thanks for fixing that.

Mobile app works properly, everything is all set to merge!

Issue #132, opened 2020-11-07
User Story 4.9


NicholasNagy:
As a stock-keeper, I want to be able to log in without a password, so that I don’t have to remember another login credential.

### Description:
This user story focuses on the PIN feature of the stock-keeper used for logging in. Once the e-mail has been entered, an expirable PIN is sent to this e-mail and has to be entered in order to access to the application.

### Acceptance Test:

1. Stock-keeper enters e-mail address corresponding to the one that was used when profile was created
2. Enters the PIN that was just sent by email
3. Log in successful by arriving on front page (list of audits page)

### Related Mockups:

- Stock-keeper login page:
> NB: when stock-keeper enters email, needs to enter PIN sent to his/her email
<img width="504" alt="sign in" src="https://user-images.githubusercontent.com/35980232/103600678-47ed5180-4f08-11eb-8711-ef41f6ffc5bb.png">

Issue #49, opened 2020-10-18
User Story 3.1


NicholasNagy:
### User Story: ALTA-3.1

As an inventory manager, I want to be able to create a stock-keeper’s user account so that I can grant them access to ALTA to perform audits.

### Description:

This feature requires that the inventory manager be able to create a stock-keeper account (which should be just enabling some permissions at this point), and subsequently building the mobile interface to allow for these users to login.

### Acceptance Test:

1. Log in as an inventory manager
2. Select the employees tab in the sidenav bar
3. Click on create, and fill in the required fields, including setting the role to stock-keeper
4. Once created, the user can then open up the mobile app, and login with the specified credentials.

### Related Mockups:

- Inventory manager login page
<img width="659" alt="log in" src="https://user-images.githubusercontent.com/35980232/103579566-c9c58680-4ed8-11eb-946e-f837b6382b7d.png">

- Side navigation bar for clicking for the employees tab (which in this case is called manage members, but in the inventory manager interface, it will be called "Employees")
<img width="659" alt="dashboard side nav" src="https://user-images.githubusercontent.com/35980232/103579607-dd70ed00-4ed8-11eb-9ad4-b7adba1331e4.png">

- The page for creating a new user
<img width="659" alt="profile" src="https://user-images.githubusercontent.com/35980232/103579629-e6fa5500-4ed8-11eb-856a-7a3e9fc2fe7a.png">

- The login page for the stock-keeper (this is mobile view)
![image](https://user-images.githubusercontent.com/40705372/101267648-f1ce9800-3728-11eb-919b-eaa0e5b5484d.png)

Amanda-Kolopa:
This issue was completed in Sprint 4.
I accidentally reopened it.

PR #165, opened 2020-11-28
Issue #49: Granting Stock Keepers access to ALTA through the mobile client

makchamp:
**New API endpoint to login from the mobile client**
- New 'login-mobile' endpoint will handle logging in exclusively from the mobile client
This new endpoint will distinguish from which client the user is logging in from
For now the authentication logic is still exactly the same (it will be changed later, as the authentication process has different requirements for stock keepers see: User Story 4.9  #132)

Current issues are mainly going to be related to routing. A big head scratcher is finding a way to keep the routing between the mobile and web client separate. This is a big deal because the default route is automatically redirected to after the app is reopened after being closed (so the web dashboard will be shown as that is set as being redirected from the default route). Also we potentially might want these mobile routes to not be accessible from the normal browser.

A nice cleaner option would be to use a multi-project workspace:https://angular.io/guide/file-structure#multiple-projects

Edit: 
I have tried to implement the multiple project solution with some success but the drawbacks were too severe to go ahead with the solution. This is mainly due to ionic still requiring the entire config files (like its own package.json) to be generated, unless the command 'ionic init' is used which transforms an existing angular project into an ionic one.
However, when attempting this, porting the current web application to the multi project structure causes weird UI bugs to surface. 
It would be possible to look into this later but for now I determined it to be not worth the time effort. 

I decided to simply generate a full isolated ionic app. This has the major benefit of ease of use and maintainability as the ionic specific code will be fully segregated. Also i recon performance and device compatibility will be improved as ionic specific elements are used everywhere. The main drawbacks is that there might be some minor code redundancies and a less elegant directory structure. 

The ionic app (''alta-mobile') now includes:
- Fully ported authentication and guards
- Implementing canLoad to secure the routes
- Log in form
- Side menu navigation

From inside the `client/alta-mobile` directory there is no need to specify the project name anymore to run any of the ionic commands:
For example:
To start the app:
`ionic serve`
To add changes to native builds:
`ionic cap copy`


NicholasNagy (approved):
I tested everything out extensively, I don't see any issues feature wise, nor code wise.
I will merge it now.

Issue #161, opened 2020-11-22
Testing/Linting Script


NicholasNagy:
Develop a script that will run our tests and linters for us easily when developing.

- The script needs to be X-Platform
- The script needs to be able to run all the tests (frontend,backend,e2e) and linters with no parameters passed to it.
- The script should allow us to pass parameters to it (using sys.argv from the python lib), to be able to run particular tests (frontend, backend or e2e) or linters (pylint or ng lint).
- Passing allTests as a parameter will get it to run all the tests (front,backend,e2e)
- Passing tests as a parameter will run only the frontend and backend tests
- Passing linters as a parameter will run only the linters.

PR #168, opened 2020-11-29
Issue #161 - Testing/Linting Script

Neeham:
Issue #161 

Added a script named **testlint.py** which can be found under the script directory.

The script can be ran using any of the following 4 commands:

1. **python testlint.py** - No arguments provided, therefore this will execute Backend test, Frontend test, E2E, and Backend/Frontend Linter
2. **python testlint.py alltests** - Will only execute the Backend test, Frontend test and E2E
3. **python testlint.py tests** - Will only execute the Backend test and the Frontend test
4. **python testlint.py linters** - Will only execute Backend/Frontend Linter


script/testlint.py
```diff
@@ -0,0 +1,119 @@
...
+            "\nArgument not recognized.\n\n"
+            "Examples to execute this scripts are:\n"
+            "python testlint.py\n"
+            "python testlint.py alltests\n"
+            "python testlint.py tests\n"
+            "python testlint.py linters")
+else:
+    print("\n The following will be executed: Backend, Frontend, E2E, Linters")
+    time.sleep(5)
+    execute("1", "1", "1", "1")
```
NicholasNagy, line 119: why are we using strings as 1's and 0's instead of actual boolean variables?

Neeham: Good catch, it honestly slipped out of my mind to make that change towards the end.

Will fix it right away.

NicholasNagy (approved):
Everything works well now, thank you for the changes!

Issue #164, opened 2020-11-28
Un-paginated inventory item API calls


NicholasNagy:
### Details
**Summary**
When retrieving inventory items from the backend, it is sending all of the organizations' items at once, and letting the front end parse and paginate the data. With millions and millions of entries, this would most likely cause major slow downs on the front-end and massive slow downs and memory use that would want to be avoided. 

This also opens up the application to DoS, because these large unnecessary I/O loads, pose a large burden on the servers that would have to provide this data.

As such, we will need to implement pagination in the backend to avoid these issues. 

**Reproduction Steps:**

While logged in as an inventory manager, go to inventory items.
If there are millions of entries of items in the database, the user should see a slow down.

**Expected Results:**
It should be quick and responsive.

**Actual Results:**
It could be potentially slow.

### Tracking
**Severity:** High
**Priority:** High
**Status:** New

PR #174, opened 2020-11-29
Pagination done in django

domludera:
Issue #164 
Refactored code such that pagination is done in the back end, this reduces the IO operation when user wants to access items information. 
Also added API calls in the front end such that it calls respective pages.


client/src/app/components/manage-inventory-items/manage-inventory-items.component.spec.ts
```diff
@@ -1,8 +1,8 @@
-import { ComponentFixture, TestBed } from '@angular/core/testing';
-import { ManageInventoryItemsService } from 'src/app/services/manage-inventory-items.service';
-import { HttpClientModule } from '@angular/common/http';
-import { RouterTestingModule } from '@angular/router/testing';
-import { ManageInventoryItemsComponent } from './manage-inventory-items.component';
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {ManageInventoryItemsService} from 'src/app/services/manage-inventory-items.service';
+import {HttpClientModule} from '@angular/common/http';
+import {RouterTestingModule} from '@angular/router/testing';
+import {ManageInventoryItemsComponent} from './manage-inventory-items.component';
```
NicholasNagy, line 5: why reformat it like this?

client/src/app/components/manage-inventory-items/manage-inventory-items.component.ts
```diff
@@ -1,20 +1,39 @@
...
+import {ViewChild} from '@angular/core';
+
+import {ManageInventoryItemsService} from 'src/app/services/manage-inventory-items.service';
+
+import {MatPaginator} from '@angular/material/paginator';
+import {PageEvent} from '@angular/material/paginator';
+
+import {MatTableDataSource} from '@angular/material/table';
+import {MatSort} from '@angular/material/sort';
+
```
NicholasNagy, line 11: Why remove the spaces and reformat it like this?

server/django_server/settings.py
```diff
@@ -41,6 +41,8 @@
     'DEFAULT_PERMISSIONS_CLASSES': (
         'rest_framework.permission.IsAuthenticated'
     ),
+    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.PageNumberPagination',
+    'PAGE_SIZE': 25,
```
NicholasNagy, line 45: Having line 44, will mess with retrieving from all the other APIs, so when I log in as a sys admin, retrieving organizations and manage members, returns no organizations or employees respectively. 
Can you set these parameter explicitly in the view so that we don't have these adverse side effects?

NicholasNagy (changes requested):
Overall, everything looks fine, except, I think we will need to give the custom behaviour to the specific view by creating a pagination class.

Here is an example of what it may look like:

```
class InventoryPagination(PageNumberPagination):
    page_size = 25
    page_size_query_param = 'page_size'
    max_page_size = 100
```

Try this out, and lmk if it works. I also left some comments related to styling.

client/src/app/components/manage-inventory-items/manage-inventory-items.component.spec.ts
domludera: Those were auto reformatted by the IDE

client/src/app/components/manage-inventory-items/manage-inventory-items.component.ts
domludera: Those were auto reformatted by the IDE

server/django_server/settings.py
domludera: I added a local class for the pagination of the items such that it doesn't affect the rest of the paginations

NicholasNagy (approved):
I looked over everything, and it looks good now, thank you for the changes!

PR #175, opened 2020-11-29
Updating System tests for 2.3 and 3.2

Averynder:
This adds the tests for 2.3 and 3.2 which adds tests to show that the IM can modify the stock keeper's information as well as that the inventory data can be accessed, currently 3.2 is set to forbidden, it should be changed to successful once the user story has been completed


makchamp (approved):
These tests were reviewed on call. 
The tests are appropriate and functional.

Issue #151, opened 2020-11-11
Fix issues raised in Acceptance Test 1.3


NicholasNagy:
• need to be able to include the organization's location upon creation and edit
• instead of deleting an organization, change status to disabled or inactive

Status - Pending

PR #177, opened 2020-11-30
Issue #151: Fix issues raised in Acceptance Test 1.3 

KhaledBadran:
This PR introduces a fix for user story 1.3. The following changes are introduced by this PR:
- new create organization dialog (added form with address field)
- new edit organization dialog (added form with address and status (active/disabled)
- removed delete option
- changed update request from put to patch
- refactored create request 
- showing address in table instead of just Montreal QC
- new organizations created with true (active) status now


makchamp (approved):
This work has been reviewed manually with @KhaledBadran and tested. Everything is working properly. Great work opting to go with a more flexible approach of templates for the edit and create organization dialogs (instead of generating them programmically which is harder to change and test)

PR #179, opened 2020-12-02
fix: Fix app compilation errror caused by old ionic styling modules

makchamp:
Issue #180 Deleted all the remaining the ionic references from the app


RobertBeaudenon:
Thanks for fixing it, I'm merging.

RobertBeaudenon (approved):

PR #181, opened 2020-12-03
Tests Fixtures Update

Averynder:
This PR does 2 things,

1. it moves the user_account permissions which were originally separated from the rest of the code in the django_server into the user_account directory / keeping all the user account code together
2. it replaces the created database entries for each test by loading the entries in the fixtures instead. This lead to a few tests being removed as they were included within another test after simplifying


Amanda-Kolopa (commented):
Everything appears to be good and running on my end.
I'll let @RobertBeaudenon have the final review before merging to double verify.
Perhaps you can fix the detected code smell in the meantime?

server/user_account/tests.py
```diff
@@ -382,69 +213,34 @@ def test_logout_failure(self):
...
-            role='SK',
-            is_active=True,
-            organization=organization)
+        self.system_admin = CustomUser.objects.get(user_name="sa")
+        self.manager = CustomUser.objects.get(user_name="im")
+        self.stock_keeper = CustomUser.objects.get(user_name="sk")
 
         self.sys_admin_id = self.system_admin.id
-        self.test_user_id = self.test_user.id
+        # self.test_user_id = self.test_user.id
```
NicholasNagy, line 229: Why are you commenting this out?

NicholasNagy (commented):
Looks good, thank you for moving things around and cleaning up the tests.

I left a single comment for discussion, before I can make a judgement to merge or not

server/user_account/tests.py
Averynder: sorry, it was a forgotten line to be removed. test_user was completely removed as it had no real purpose and it's attributes were the same as admin. Before deleting code I usually comment it out to verify nothing else is being used or if I want to go back to this code for whatever reason. Good catch!

NicholasNagy: np!

NicholasNagy (approved):
Everything looks good now, I will merge it!

Issue #48, opened 2020-10-18
User Story 2.5


NicholasNagy:
### User Story: ALTA-2.5

As an inventory manager, I want to be able to search for an inventory item through ALTA so that I can include it to a potential audit.

### Description:

The user should be able to search the inventory based on the characteristics (fields) of the items.

### Acceptance test ():

1. Log in as an inventory manager.
2. On the side navigation bar, click on the _Inventory_ tab.
3. At the top of the list of items, there should be some functionality related to searching that can help the user search for various fields.

### Related Mockups:

- Inventory manager login page
<img width="659" alt="log in" src="https://user-images.githubusercontent.com/35980232/103579566-c9c58680-4ed8-11eb-946e-f837b6382b7d.png">

- Inventory page has a search bar for specific fields
<img width="659" alt="Capture d’écran 2021-01-05 à 03 36 56" src="https://user-images.githubusercontent.com/35980232/103600320-669f1880-4f07-11eb-8539-b7b15feae488.png">

<img width="659" alt="inventory fields" src="https://user-images.githubusercontent.com/35980232/103806642-a07e3500-5055-11eb-8c52-831f95d61112.png">

PR #186, opened 2020-12-13
Filtering and Searching of items

domludera:
This PR introduces filtering and searching for items in the dashboard
User Story 2.5 Issue #48 

The front end generates a parameterized URL based on what the user has entered in the "Search Item" menu and requests the items from the back end.


Averynder:
Please remove / make the same for part description, criticality and plant and also we receive an error if on the 2nd page when hitting new submissions

Averynder (changes requested):
Please add the above changes

Averynder:
Functionality works well, refactor the inventory view to search through sent arguments rather then checking switch based for the same one.
Ex: https://stackoverflow.com/questions/3294889/iterating-over-dictionaries-using-for-loops

Also fix the linting issues

Averynder (approved):

PR #200, opened 2021-01-09
BugFix: Side Nav missing Information

Averynder:
Previously, the sidenav was missing fields as they were not being returned in the serializer's defaults. This wasn't an issue previously because the requested fields were sent in the prior commit however since removing them as of the refactored view changes they were incomplete / missing the required fields desired by the client

### DIFF
--- a/client/src/app/components/create-member/create-member.component.ts
+++ b/client/src/app/components/create-member/create-member.component.ts
@@ -1,9 +1,9 @@
-import { Component, OnInit } from '@angular/core';
-import { AuthService } from 'src/app/services/auth.service';
-import { FormGroup, FormBuilder, Validators } from '@angular/forms';
-import { Router } from '@angular/router';
-import { TokenService } from 'src/app/services/token.service';
-import roles from 'src/app/fixtures/create_members_roles.json';
+import {Component, OnInit} from '@angular/core';
+import {AuthService} from 'src/app/services/auth.service';
+import {FormGroup, FormBuilder, Validators} from '@angular/forms';
+import {Router} from '@angular/router';
+import {TokenService} from 'src/app/services/token.service';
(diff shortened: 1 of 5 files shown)


client/src/app/components/sidenav/sidenav.component.ts
```diff
@@ -42,6 +42,7 @@ export class SideNavComponent implements OnInit {
         this.loggedInUser = data.username;
         roles.forEach(role => {
           if (role.abbrev === data.role) {
+            this.loggedInUser = data.username;
```
KhaledBadran, line 45: This line is a duplicate of line 42. I believe that it should be removed.

client/src/app/components/create-member/create-member.component.ts
```diff
@@ -3,7 +3,7 @@ import { AuthService } from 'src/app/services/auth.service';
 import { FormGroup, FormBuilder, Validators } from '@angular/forms';
 import { Router } from '@angular/router';
 import { TokenService } from 'src/app/services/token.service';
-import roles from 'src/app/fixtures/roles.json';
+import roles from 'src/app/fixtures/create_members_roles.json';
```
KhaledBadran, line 6: I think it is better to change the name of the import to be create_member_roles:

`import create_members_roles from 'src/app/fixtures/create_members_roles.json';`

and to rename the `roles` variable as `create_members_roles` in the component and html files so that it doesn't get mixed up with the `roles` variable used in other components.

KhaledBadran (changes requested):
Thank you @Averynder for the bug fix 👍. The fix works well and the code functions as expected. I requested some minor changes to the code in order to for the pr to be approved.

KhaledBadran:
The required fixes have been pushed. It all looks good now 👍

KhaledBadran (approved):

Issue #52, opened 2020-10-18
User Story 3.4


NicholasNagy:
### User Story: ALTA-3.4

As an inventory manager, I want to be able to modify a customizable auditing template so that I can adjust the specified requirements of a potential audit.

### Description

Following the same concept as user story 3.3 #51 , when logged in as an inventory manager, clicking on the template tab in the sidenav should bring the user to the list of templates. Clicking on any one of them should take you to the form of the template for which you can edit the information. This assumes that 3.3 is finished, so you can most likely re-use the same component.

### Acceptance Test

1. Log in as an inventory manager
2. Click on the Templates tab in the side nav
3. Click on any one of the templates to edit it.
4. The app should display the template with the fields greyed out, and a button to begin editing
5. Once clicked, the button should ungrey the fields and allow the user to edit the information.
6. Upon completion of the edit, the user should have a submission button to submit the edits.
7. Upon submission, the information should be updated in the database

### Related Mockups

- Inventory manager logs into ALTA
<img width="617" alt="log in" src="https://user-images.githubusercontent.com/35980232/103582973-2f1c7600-4edf-11eb-966b-eb8f33e2a7e8.png">

- Inventory manager is presented with the Dashboard page
<img width="617" alt="dashboard side nav" src="https://user-images.githubusercontent.com/35980232/103582990-3479c080-4edf-11eb-8aac-357958c86f2f.png">

- Navigate to the Templates page and select a template
<img width="617" alt="templates" src="https://user-images.githubusercontent.com/35980232/103583002-3a6fa180-4edf-11eb-8601-6d21cb6d5c8e.png">

- Display the template form with the fields containing predefined data
<img width="617" alt="Template edit" src="https://user-images.githubusercontent.com/35980232/103761380-dcde7080-5016-11eb-9503-52243f3725e9.png">

- Edit and save the updated data
<img width="617" alt="template save" src="https://user-images.githubusercontent.com/35980232/103761394-e2d45180-5016-11eb-936e-b3bc4f734a88.png">

PR #201, opened 2021-01-09
Edit template

NicholasNagy:
This issue aims to complete the functionality for user story 3.4 (#52).

The intention is to allow users to edit created templates for their own organization (including deletion).

The permissions are also intended to be set up to prevent users from modifying templates and retrieving templates from other organizations.

### DIFF
--- a/client/src/app/components/audit-template/create-audit-template/create-audit-template.component.html
+++ b/client/src/app/components/audit-template/create-audit-template/create-audit-template.component.html
@@ -26,7 +26,7 @@
             <mat-label>Location</mat-label>
             <mat-form-field appearance="outline">
               <input class="form-control" id="location" type="text" matInput [(ngModel)]="templateValues.location"
-                     placeholder="Location"/>
+                     placeholder="Location" [disabled]="disabled"/>
               <button mat-icon-button matSuffix (click)="addItem('location', templateValues.location)">
                 <mat-icon>add_circle_outline</mat-icon>
               </button>
@@ -37,7 +37,7 @@
             <mat-label>Plant</mat-label>
             <mat-form-field appearance="outline">
               <input class="form-control" id="plant" type="text" matInput [(ngModel)]="templateValues.plant"
-                     placeholder="Plant"/>
+                     placeholder="Plant" [disabled]="disabled"/>
               <button mat-icon-button matSuffix (click)="addItem('plant', templateValues.plant)">
                 <mat-icon>add_circle_outline</mat-icon>
               </button>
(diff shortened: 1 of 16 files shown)


client/src/app/components/audit-template/edit-audit-template/edit-audit-template.component.ts
```diff
@@ -0,0 +1,147 @@
...
+
+interface Template {
+  location: any;
+  plant: any;
+  zones: any;
+  aisles: any;
+  bins: any;
+  part_number: any;
+  serial_number: any;
+}
```
Amanda-Kolopa, line 14: Can this be moved to its own model.ts?
We could reuse it

client/src/app/components/audit-template/create-audit-template/create-audit-template.component.html
```diff
@@ -126,9 +126,13 @@
     </mat-card-content>
   <mat-card-footer>
     <div class="center-aligned">
-      <div class="button-side-by-side">
+      <div class="button-side-by-side" *ngIf="disabled">
+        <button mat-stroked-button id="edit-template-button" class="save-button-bg" (click)="beginEdit()">
+          EDIT</button><!-- if checks throws error then do not send data -->
+      </div>
```
Amanda-Kolopa, line 132: Can we remove the Edit button since the component manages audit creation?
I really don't like that it calls an empty function

client/src/app/components/audit-template/create-audit-template/create-audit-template.component.ts
```diff
@@ -110,4 +110,8 @@ export class CreateAuditTemplateComponent implements OnInit {
     }
 
   }
+
+  beginEdit(): void {
+    // do nothing
+  }
```
Amanda-Kolopa, line 116: This should be removed since it does nothing

Amanda-Kolopa (changes requested):
Overall great work!

Some minor changes:

- When the user selects an audit to edit, the _Selected Fields_ items display an 'X' signifying that it can be removed. The 'X' should only display once the user selects the bottom 'Edit' button

client/src/app/components/audit-template/create-audit-template/create-audit-template.component.html
NicholasNagy: In order to avoid HTML duplication, I need to keep this button here so that I could use it with the edit template component.

NicholasNagy:
Hey @Amanda-Kolopa , I made the fixes, please also see my reply to your second comment for why I can't simply remove that button in the creation page.

Amanda-Kolopa (approved):
This PR is good to merge :smile:

PR #205, opened 2021-01-14
[U4.3 & U.4.4] Bacode scanning with external bluetooth scanner, camera and manual input

makchamp:
##### To start localtunnel for the django server :
`npx localtunnel --port 8000 --subdomain {CUSTOM_DOMAIN}`
The custom domain provided (or generated if none is provided) needs to be added to both environment files
- For the ionic app, set `api_root` to the custom domain
- In the .env file for the django server, set `HOST_URL` to the custom domain
- Localtunnel is used as android only allows HTTPS requests by default.
- The tunnel needs to be restarted if left open for a very long time

##### Setup npm scripts for live reloading of android and ios apps
To start the live reload server for the mobile apps we can now run the commands:
`npm run live-android` or `npm run live-ios`
This will generate a url that will have to be added to the .env file of the django server (for `ANROID_URL`and/or `IOS_URL`)
This is very useful to speedup development.

### Barcode scanning:
- Setup Audit service and model (not actually used yet, I hardcode a dummy list of items for now)
#### Scanning using an external bluetooth scanner
- Listening to keypress events on the Audits page (The Bluetooth scanner sends input (as a normal keyboard would) when a barcode is being scanned)
- A top loading(Scanning...) message is displayed at the top of the screen (below the toolbar) when scanning is taking place. This includes a ion-progress-bar.
- The scanned barcode is saved as a string and is then validated with the list of items (will be replaced with backend request)
#### Scanning using the camera
- If the bluetooth scanner is lost or not available, the camera can be used for scanning
- This is done with the help of 'Barcode Scanner' ionic plugin
- On android devices, the prompt to allow camera permissions will only be shown when the user requests to scan with the camera (https://ionicframework.com/docs/native/android-permissions)
#### Manual input:
- If both methods listed above are unavailable, the user can enter the barcode # manually. 
- This is done through an alert prompt
- This can later be expanded to query items using more information than the barcode # alone.


Averynder (changes requested):
Looks good and runs smoothly, does exactly as specified, a few notes to consider:
- Hot reload wasn't able to load
- This was tested on android for the time being
- alta mobile should be it's own specific folder similar to the client and server

makchamp:
> Looks good and runs smoothly, does exactly as specified, a few notes to consider:
> 
>     * Hot reload wasn't able to load
> 
>     * This was tested on android for the time being
> 
>     * alta mobile should be it's own specific folder similar to the client and server

- Live reload might not have worked for you due to the network settings on your VM blocking the eternal live reload server or the config url being wrong or something else entirely that I would not be aware of at the moment. 
- Yes I am currently working on getting a mac environment up and running. 
- Yes I agree, I moved it into its own dir `mobile` in project root.

Issue #206, opened 2021-01-16
ID for Audit Template is not generated


domludera:
### Details

**Summary**
When creating an Audit Template, the content is passed to the database, however the collection doesn't auto create a template_id for the template. Thus it is not possible to edit nor delete the template. 
This bug isn't consistent across different computers.

**Reproduction Steps:**
Have Mongo database running
Run Makemigrations
Migrate the newly generated model
Generate new Audit Template
Try to edit or delete the newly generated template

**Expected Results:**
The request will be made to the back end and it will be processed depending on what was requested. If edit, the app will let the user edit the template. If delete, the app will delete the chose template.

**Actual Results:**
The back throws an exception as field 'template_id' expected a number but got 'null'

### Tracking

Severity: High
Priority: Low
Status: New

PR #207, opened 2021-01-16
Change Auto generated template_id field to an UUIDField

domludera:
Issue #206 

Change template_id field type from AutoField to UUIDField. This fixes the issue where template_id wouldn't be auto-generated in certain databases.


NicholasNagy (approved):
Thank you for the quick fix.
Just to document what we discussed over the call, thank you for putting back the migrations and adding the dependency to the requirements.txt.
And finally thanks for fixing the tests to add the UUID to replace the previous integer value for the ID.

The PR is good to merge now, thanks for fixing the issue!

Issue #53, opened 2020-10-18
User Story 3.5


NicholasNagy:
### User Story: ALTA-3.5

As an inventory manager, I want to be able to assign stock-keepers to perform audits so that I can manage the execution of multiple tasks simultaneously.

### Description

After an audit has been initiated the user should be taken to a screen where they will assign which stock-keepers will be assigned to the audit. At this point, it should only store this information in the database.

### Acceptance Test

1. Log in as an inventory manager
2. There should be 2 ways to initiate an audit
      a. When looking through the inventory items, the user should be able to select multiple items and initiate an audit
      b. From the templates page, the user should be able to select a template to initiate an audit for.
3. Once an audit has been initiated, it should take the user to a page where they can select stock-keepers (which is divided by location, if applicable)
4. Once stock-keepers are selected, the user should be prompted to submit through a button, which can allow the inventory manager to drag and drop each bin to each stock-keeper to assign their workload (only applicable for locations where more than 1 stock-keeper was assigned).
5. Upon submission of the workload, the user is either taken to the next workload page, where there will be more bins to assign to stock-keepers if there are more locations with more than 1 stock-keeper assigned, otherwise, the information will be submitted to the database.

### Related Mockups

- Inventory manager logs into ALTA
<img width="630" alt="log in" src="https://user-images.githubusercontent.com/35980232/103587683-106ead00-4ee8-11eb-8e3c-87123f54feb8.png">

- Dashboard once logged in shows the top row with 'Initiate audit', once clicked it brings this pop-up window to choose the 2 different way to initiate an audit: 1) From predefined template 2) From inventory items
<img width="630" alt="dashboard" src="https://user-images.githubusercontent.com/35980232/103588206-f71a3080-4ee8-11eb-9e3e-bcd92d2f6544.png">
<img width="630" alt="2 ways to initiate audit" src="https://user-images.githubusercontent.com/35980232/103588263-1913b300-4ee9-11eb-8a8f-9e01667166f6.png">

- 1) From predefined template
> Create or select a template
<img width="630" alt="templates" src="https://user-images.githubusercontent.com/35980232/103588381-5d06b800-4ee9-11eb-8324-6d955700d398.png">

> View template specifications
![Capture d’écran 2021-01-14 à 18 53 43](https://user-images.githubusercontent.com/35980232/104662933-d6e13300-5699-11eb-811c-a30c87754bcb.png)

- 2) From inventory items
[...]

PR #209, opened 2021-01-17
[ALTA-3.5] Assign stock-keepers to an audit

Amanda-Kolopa:
### Related issue: ALTA-3.5 (#53 )

Implemented functionalities:

- Once an audit is initiated through the _Inventory Items_ page, the user can select stock-keepers from several locations
- Next, the user can designate bins to certain stock-keepers through a drag and drop list
- When designation is complete, the audit components are added to the database using the audit_id as a reference


NicholasNagy (changes requested):
I think the overall functionality is really nice. My only request is that we get the drag and drop when assigning an sk to shrink in width on a small screen (I can walk you through what Neeham did for the template audits or you can use css grid boxes).

Also, when I was testing out the features on my end. I was getting failures in the permissions when trying to assign stock keepers to locations and subsequently assigning bins to each stock keeper. As a result I had to loosen the permissions and simply return true. Please have a  look.

NicholasNagy:
Thanks for all the changes, everything looks good now!

NicholasNagy (approved):

PR #210, opened 2021-01-20
Epics 1 and 2 UI tested

Averynder:
This PR now has epics 1 and 2 with the exception of user story 2.4, all of which are now tested via the e2e tests. Comments have been added to detail where each test has came from along with what each test is doing in verbal terms rather then having the dev attempt to figure out what the test is doing during run time.

It should be noted that the original e2e script was used and that the current testlint.py requires a modification to be run. This is already fixed within the performance testing branch


Amanda-Kolopa (changes requested):
Thank you for the great work! :smile: 

Some notes:

- I experienced some issues when trying to run them since Chrome recently released an update that my browser didn't pick up on right away

  - To fix:
     - Select browser menu (3 dots) > Help > About Google Chome
     - The update should start then

  - If that doesn't work: 
    - download [ChromeDriver 88](https://sites.google.com/a/chromium.org/chromedriver/downloads) 
    - unzip file
    - update path environment variable to unzipped .exe location
    - navigate to your Chrome location and remove the older version folder from the directory 

- A couple of tests are failing on my end with this error (specifically, 1.5/1.6 and 2.2)
   ```
   Failed: javascript error: document unloaded while waiting for result
   ```
  - I looked into it and added some sleep statements to give it more time to check for a result. Please run the tests on your end to ensure the fixes are alright!

- SonarCloud mentions security hotspots but these are false positives since it picks up 'password' in plaintext. Perhaps we can suppress these warnings by trying these: https://docs.codescan.io/hc/en-us/articles/360012109711-Ignoring-violations

- Please fix the code based on the lint reports and code smells. Also, if you can look into removing some duplicate code, it would be great!

Amanda-Kolopa (approved):
Thanks for the changes! Good to merge.

PR #211, opened 2021-01-24
Role creation bypass

Averynder:
This PR fixes the previous security vulnerability discovered where IMs could create SAs

### DIFF
--- a/server/user_account/permissions.py
+++ b/server/user_account/permissions.py
@@ -142,3 +142,18 @@ def check_keys(bad_keys, keys):
         if bad_key in keys:
             return False
     return True
+
+
+class CanCreate(BasePermission):
+    message = "Only System Admins can create other System Admins"
+
+    def has_permission(self, request, view):
+        """
+        :param request: Getting the user that is doing the request
+        :param view: Getting the targeted pk passed in the URL
+        :return: True/False : Whether the user is allowed to create the user
+        """
+        if IsSystemAdmin.has_permission(self, request, view):
+            return True
+        if IsInventoryManager.has_permission(self, request, view):
+            return not request.data['role'] == 'SA'
--- a/server/user_account/tests.py
+++ b/server/user_account/tests.py
@@ -145,6 +145,13 @@ class RegistrationTestCase(APITestCase):
                                    format='json')
         self.assertEqual(request.status_code, status.HTTP_403_FORBIDDEN)
 
+    def test_im_creates_sa(self):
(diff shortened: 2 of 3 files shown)


server/user_account/permissions.py
```diff
@@ -142,3 +142,18 @@ def check_keys(bad_keys, keys):
...
+    def has_permission(self, request, view):
+        """
+        :param request: Getting the user that is doing the request
+        :param view: Getting the targeted pk passed in the URL
+        :return: True/False : Whether the user is allowed to create the user
+        """
+        if IsSystemAdmin.has_permission(self, request, view):
+            return True
+        if IsInventoryManager.has_permission(self, request, view):
+            return not request.data['role'] == 'SA'
```
NicholasNagy, line 159: Can't you just add this logic to the IsInventoryManager permissions? I feel like adding this entire class for that one check is a bit excessive

NicholasNagy (commented):
Please see my comment, otherwise, it's good, thank you for the quick fix

server/user_account/permissions.py
Averynder: > 
> 
> Can't you just add this logic to the IsInventoryManager permissions? I feel like adding this entire class for that one check is a bit excessive

So it could be added there but then we would be going against what the initial permission is doing which is simply checking if the user is an IM. This permission is used elsewhere and the code wouldn't apply there then, say for example in Audits. The way it is now, this permission is only used when trying to Create and no where else otherwise

NicholasNagy: I agree, but can't we reuse this code for the patch method? Wouldn't the logic be exactly the same?

Averynder: > 
> 
> I agree, but can't we reuse this code for the patch method? Wouldn't the logic be exactly the same?

The way it is now it isn't. In the patch method how it's checked is based on keys you shouldn't have, role is popped before it ever sees the serializer since roles can never be changed regardless if you are an SA. In this case, the role could or couldn't not be an SA depending on who's making the call which is why we have a separate permission. In the Patch method, we have to send the role though because we end up comparing it with the person being 'patched' to see if the operation is permitted and then to also compare the organizations as well. It's not a small change unfortunately, otherwise i agree they should be merged, functionality wise they're not much different from a dev perspective but from the requirements we have in our system it becomes more complex though not impossible

NicholasNagy:
This was closed because the issue was not resolved in this PR.
A new PR will be opened soon to fix the issue

Issue #212, opened 2021-01-24
IM bypass security by creating SA


Averynder:
### Details

**Summary**
Currently an inventory manager can create an SA by submitting a request to the API bypassing the entire system's security

**Reproduction Steps**:

1. The user logs in as an IM
2. The user submits a POST request to /user/ and replaces role with 'SA'

**Expected Results**:
The request will be blocked with error 403

**Actual Results**:
The request will go through and return the newly created SA's profile

### Tracking

Severity: High
Priority: High
Status: Fixed

PR #213, opened 2021-01-24
Refactor create edit

NicholasNagy:
This PR refactors 3 major areas in our application.

1. The audit template components for creating and editing
2. The organization components for creating and editing
3. The employee/profile settings components for creating and editing user profiles

The driving factor for this refactoring is that there are many areas in which there are code duplicates of fields, methods, and sometimes functionality. This is as a result of the HTML components that are either very similar (in the case of the employee/profile settings components) or that they use the same component. 

The new structure is as follows:

The 2 previous components (for creating and editing) now will inherit similar functionality from a super class. Each of the components will also reuse the same HTML file, and the same SCSS file for displaying and styling the information.


client/src/app/components/audit-template/audit-template-view/audit-template-view.component.ts
```diff
@@ -0,0 +1,75 @@
...
+  ngOnInit(): void {
+    this.initializeForm();
+  }
+
+  abstract initializeForm(): void;
+
+  addItem(term: string | number, value: string): void {
+    console.log("hello");
+    console.log(term);
+    console.log(value);
```
Amanda-Kolopa, line 37: Please remove these

Amanda-Kolopa (changes requested):
Thank you for cleaning up the code with these refactorings! :smile:

Some notes:

- For Organizations: 

   - can you please adjust the _Organizations_ table layout as the _Status_ column's right padding squishes the _Settings_ column
image

  - when I try to create a new organization and submit an empty name, the error "An organization with this name already exists"

- For Employees:

  - When I try to create a new employee for an organization, the form allows me to submit a new user with the drop-down menu option set to "Choose a role".  The backend accepts the request and sets the newly created user's role to SA. This happens when I log in as an SA and IM.

- For Templates:

  - When I edit a template, the fields' + buttons are clickable. Please disable these until the _Edit_ button is selected

- I noticed in manage-organizations\organization-list there is organization.dialog.html that I don't think it's being used anymore by the component. At the bottom of manage-organizations.components.ts, the class is there but appears to not hold much value. Can you confirm if we still need this dialog box, please?

NicholasNagy:
> Thank you for cleaning up the code with these refactorings! 
> 
> Some notes:
> 
> * For Organizations:
>   
>   * can you please adjust the _Organizations_ table layout as the _Status_ column's right padding squishes the _Settings_ column

done

>   * when I try to create a new organization and submit an empty name, the error "An organization with this name already exists"

fixed

> * For Employees:
>   
>   * When I try to create a new employee for an organization, the form allows me to submit a new user with the drop-down menu option set to "Choose a role".  The backend accepts the request and sets the newly created user's role to SA. This happens when I log in as an SA and IM.

I cannot replicate this, if you are talking about #212 , it is still a work in progress.

> * For Templates:
>   
>   * When I edit a template, the fields' + buttons are clickable. Please disable these until the _Edit_ button is selected

Fixed

> * I noticed in manage-organizations\organization-list there is organization.dialog.html that I don't think it's being used anymore by the component. At the bottom of manage-organizations.components.ts, the class is there but appears to not hold much value. Can you confirm if we still need this dialog box, please?

Good catch, I remember looking at it and putting it on my mental to-do list, but forgetting!

NicholasNagy:
> > * For Employees:
> >   
> >   * When I try to create a new employee for an organization, the form allows me to submit a new user with the drop-down menu option set to "Choose a role".  The backend accepts the request and sets the newly created user's role to SA. This happens when I log in as an SA and IM.
> 
> I cannot replicate this, if you are talking about #212 , it is still a work in progress.

We ended discussing and @Amanda-Kolopa ended up showing me what she meant. Basically, it's that if no role was assigned, the front end would automatically assign the role to stock-keeper, and that the button to submit the form wasn't disabled when the form was invalid.

Amanda-Kolopa (approved):
This PR looks good to merge :slightly_smiling_face:

Issue #224, opened 2021-01-26
Side Navigation Menu Item Not Being Highlighted After Page Refresh


makchamp:
#### Summary
Side Navigation Menu Item Not Being Highlighted After Page Refresh
#### Reproduction Steps:
- Log in as any user
- Go to Add Template or Create/Modify Member page
- Refresh page

#### Expected Results:
Relevant menu item is still highlighted in the side nav menu
#### Actual Results:
Menu item is no longer highlighted in the side nav menu

#### Tracking
Severity: Low
Priority: Low
Status: Not fixed

PR #225, opened 2021-01-26
Issue #224 Side navigation item highlight fix

makchamp:
#### Fix for #224
The issue was related to incorrect types being used. We need to use the path (string) not the router object. I also changed the condition to use includes and not direct equals.

This fixed the issue however the new assign and designate sk pages weren't linked to any parent routes from the side nav so no item would be highlighted when accessing those particular pages.
I decided to link them to the audits parent URL like this `audits/assing-sk/designate-sk/`
The `Audits` menu item redirected nowhere so it now routes to the `Inventory Items` page temporarily.


NicholasNagy (approved):
Hey, thanks for the fix,
I don't see any issues with the code, so the this PR is good to go.

PR #234, opened 2021-01-28
Issue #53 - fixed failing tests

Neeham:
#53 - Previously the tests were created incorrectly causing a failure, however when running multiple tests the incorrect test is specified. This fixes it by having all the components created properly.


Amanda-Kolopa (approved):
Thank you for fixing the tests!
Aside from the failing SonarCloud (which is fixed in pull request #208), all is good to merge :smile:

PR #235, opened 2021-01-30
Fix permissions and permissions refactor.

NicholasNagy:
#212 This PR is opened to fix issue 212.

In addition to the fix, this PR aims to refactor our overly complex permissions. The goal was to break apart every permission that was achieving more than 1 goal, greater than what is described in its class name. This will help us when we need to make changes in our permissions, as we won't have to tack on new functionality at a random location, and pray that it works with everything else.

I also removed the dependencies on the IsInventoryManager permissions that were previously created, as any further changes in IsInventoryManager could easily affect the subclasses indirectly without our knowing. 

The permissions are now very reusable, and I've developed a common function for which most components use currently, and will most likely be using further in the future.


Amanda-Kolopa (changes requested):
Everything is working correctly for SA and IM users. Great work! :smile:
Before merging, can you please add some comments to the modified permissions and fix the code based on the pylint report?

Amanda-Kolopa (approved):
All looks good to merge :slightly_smiling_face:

Issue #55, opened 2020-10-18
User Story 3.7


NicholasNagy:
### User Story: ALTA-3.7

As an inventory manager, I want to be able to review an audit’s selected components so that I can validate the task before initiating it.

### Description:

For this story, the goal is to provide a better insight into how busy the stock-keepers and display that when an inventory manager is assigning audits. This will include displaying additional information in the interface where the inventory manager is choosing the associated stock-keepers for the audit.

**Note:** We discussed and currently the idea is to provide a high level overview by showing the number of audits assigned, but we were also thinking of adding a drill down chart with more information, such as # of bins, # items to audit to help give a better idea of the workload of the stock-keepers

### Acceptance Test:
_Considering that user is signed is as Inventory Manager_

1. Once logged in, arrive on the inventory login page
2. Initiate the audit from the _Initiate New Audit_ button
3. Select either from _Preferined Template_ or _Inventory Items_
4. Select stock-keepers per location for that new audit, click on _Initiate_ button
5. Drag bins to specific stock-keepers per location, click on _Assign_ button
6. Arrive on the _Audit Review_ page to have a recap of what has been chosen so far, click on _Initiate_ when review done

### Related Mockups:

- Inventory manager login page
<img width="659" alt="log in" src="https://user-images.githubusercontent.com/35980232/103579566-c9c58680-4ed8-11eb-946e-f837b6382b7d.png">

- Initiate audit page: assigning stock-keeper per location
![Capture d’écran 2021-01-24 à 13 26 03](https://user-images.githubusercontent.com/35980232/105639687-b65a6b00-5e47-11eb-8aa0-5370542d293f.png)

- Assigning specific bins to stock-keepers
![Capture d’écran 2021-01-24 à 13 26 17](https://user-images.githubusercontent.com/35980232/105639699-be1a0f80-5e47-11eb-9c8a-3a25771b046f.png)

- Review page 
![Capture d’écran 2021-01-24 à 13 26 27](https://user-images.githubusercontent.com/35980232/105639703-c3775a00-5e47-11eb-981a-30a7a76dca4c.png)

PR #236, opened 2021-01-30
Review/Recap audit page 3.7

daphneaugier:
PR related to user story 3.7 (#55 ): _As an inventory manager, I want to be able to review an audit’s selected components so that I can validate the task before initiating it._

What was done : 

- **Added review-audit component to have this extra page at the end of an audit to review what has been assigned to whom** 
> - Created review-audit component
> - Contains a table per location, where each table in one location is related to a stock-keeper
> - Table information is: Stock-Keeper's name, Bin, Number of parts, the date when the audit has been initiated, name of IM who initiated the audit
- Added 'GO BACK' button in order to go back and forth in between the "Assign audit to stock-keeper" phases
> - Button available for the *Select Stock-Keeper*, *Designate Bins* and *Review Audit* pages
> - Button is a part of the review feature, where when an IM clicks on it, it brings back to the previous if IM forgot something and can add it on top of currently saved information for audit creation


makchamp (changes requested):
The change is functional and does as intended. There are just a few issues that need to be addressed before this is merged:
1) There is a bug where if one clicks the back button to go back from the new Review Audit, going to the page once again will cause there to be duplicate rows in the review table. A possible reason for this is because we are not patching an already existing item but posting again. A solution to this would be to either patch an already existing entry or delete the temporary entry entirely before we go back. 
2) Some columns on the table are not aligned properly if the number of bins is not even across rows.
3) Missing fields initiated by and initiated date
Screenshot from 2021-01-30 18-16-45

daphneaugier:
Thank you for your comments Makram! 

I fixed your comment 2) from your review in my latest commit. As you told me on Discord, you took charge on fixing comments 1) and 3).

- [x] Teamwork

makchamp (approved):
Changes since the initial review:

- Changed ItemToSK model to have id field
- Changed Audit and ItemToSK viewsets to allow delete
- Change manage audit service to have delete request method
- Discard button will now delete the audit and reroute to the audit page
- Back button has been linked in all assigned audit pages, will delete itemSKData if backing from review page
- Routing with replaceURL in all the assign audit pages 
- Overriding browser back button to be the same behaviour as the custom back button
- Changed colour of discard button to red
- Link review audit page to the app navigation
- Fix sonar complaints
- Fix columns being misaligned when the number of bins is not even across rows
- Fix duplicate location containers being displayed when multiple stock keepers from the same location are assigned

The initial flagged issues have been resolved. For the 'initiated by' and 'initiated by' columns missing data, that addition can come later as it requires a more significant modification to the Audit models which is already underway anyway for the following stories.

PR #238, opened 2021-01-31
Fix item refresh im bug

Averynder:
Previously, when an IM attempted to update the job refresh time, they were unable to achieve this due to a coherency error where the permission and the view required 2 different data types, this has now been fixed


KhaledBadran (approved):
I tested the fixed functionality and the tests. All looks good 👍🏻 Thanks @Averynder and @NicholasNagy for the fixes.

PR #247, opened 2021-02-15
Centralize local storage manipulation

NicholasNagy:
This PR moves all the local storage database accesses to central locations/the related services of the application.

It also reorganizes the services into directories to make them easier to find.


Averynder (approved):
Looks good, thank you for your work, all the tests pass

PR #265, opened 2021-02-26
Updating mobile work flow

NicholasNagy:
Some of the changes in this PR include:

- fixing the process for flagging (flagging != missing item)
- When item is serialized, add a toggle (since the value can only be 1 or 0 for the quantity)
- fixed bug in getting the bin info (permissions were failing because of bad query)
- removed the circles next to bins and audits
- submitting an audit is now done automagically when the last bin is submitted
- Fixed some vulnerabilities in the permissions


makchamp (approved):
Thanks for making these tweaks. I went over all the functionality to make sure everything was working correctly. I just renamed some methods and cleaned up some code linting errors. 
There are just 2 things that come to my mind. The item id and Batch number fields are stored as a CharField in the database now but the corresponding input type is 'number' in the record modal input field(and in the manual input field). This shouldn't be an issue if we never except characters to be part of a batch number or item id, however if ever a barcode containing characters is scanned, it will not be invisible in the id input field unless its type is changed  to 'text'. The next thing is that Batch_Number is not shown in the expandable view along side the other item information fields. I do not think that is an issue as it is a little redundant anyway.

NicholasNagy:
Thanks for the feedback @makchamp as long as we don't have any bugs, I'm cool with that.
I stored it in that way in the case that these numbers get mixed up the characters, and having to debug that later could be an issue.

Issue #56, opened 2020-10-18
User Story 4.1


NicholasNagy:
As a stock-keeper, I want to be notified once an audit is assigned to me so that I can perform the audit as soon as possible.

### Description:
When logged in through the stock-keeper's interface, the user arrives on the front page which is the list of audits page. If a new audit has been assigned, the user will automatically see it through both the notification bell having a red dot and in the list of audits, the audit in question will be written in bold and have a red dot. 

### Acceptance Test:

1. Login as stock keeper using PIN
2. Arrives on the List of Audits page
3. Newly assigned audit is written in bold and has a red dot next to it, as well as the notification bell as a red dot on it

### Related Mockups:

- Stock-keeper login page:
> NB: when stock-keeper enters email, needs to enter PIN sent to his/her email
<img width="504" alt="sign in" src="https://user-images.githubusercontent.com/35980232/103600678-47ed5180-4f08-11eb-8711-ef41f6ffc5bb.png">

- Arrives on the list of audits page where the bell has red dot if an audit has been assigned. 
> NB : On the list, can easily see which audits are new as they are in bold and have the same red dot as the one on the bell
![notification](https://user-images.githubusercontent.com/35980232/104663831-d9dd2300-569b-11eb-9870-a026fc341ff0.png)

Amanda-Kolopa:
This user story was completed in pull request #253

PR #267, opened 2021-03-06
U4.1 - Mobile Audit notifications

makchamp:
#### This PR covers U4.1 #56 

Added an Assignment relation to the Audit model. This will provide additional information specific to each individual stock keeper's audit assignment (like if they have seen/initiated the audit or not). This relation is similar to the ManytoMany field assigned_sk from the Audit model. The assigned_sk ManytoMany table is expandable with the 'through' keyword however the feature is not very stable and caused a few new issues. The explicit Assignment relation makes that field redundant but it was kept as it is currently referenced many times in the app.
Implemented an Assignment viewset similar to the Audit viewset. 

On the mobile app, the user will now receive in-app notifications anytime an audit is assigned to them. Initially an audit assignment is marked as unseen. All unseen audit assignments will show up in a popover list after clicking the bell notification icon from the toolbar. The number of new notifications appears next to the bell icon. New/unseen audits will also appear in bold in the main list view with a small red icon next to them. An unseen audit is marked as seen as soon when the user first enters/clicks on it.


Averynder (approved):
Looks like a good PR, code was reviewed and produces the desired effect.

PR #268, opened 2021-03-11
Autocomplete search results for the Audit Templates Fields

makchamp:
On the add audit template page, the following input fields now support autocomplete results: 
- Location
- Plant
- Zone
- Aisle
- Part_Number
- Serial_Number
Note: Aisle will not currently display anything as the django filtering only supports char types and Aisle is an integer field. 

The filtering query has been changed to use  icontains (matches substrings instead of exact matches).
The top 6 results will show in the dropdown. Results are filtered automatically as the user types in characters. Blank input will still show 6 results.
Built using the MatAutoComplete angular module: https://material.angular.io/components/autocomplete/overview


NicholasNagy (approved):
This is a relatively small change, and I don't really see anything wrong with it.

Thank you for the changes!

Issue #195, opened 2020-12-23
User Story 5.7


Amanda-Kolopa:
### User Story: ALTA-5.7
As an inventory manager, I want to be able to view a dashboard of recent audit activity so that I can review updates in a centralized location.

### Description
The purpose of the dashboard is to be the main page of the application when logged in as an inventory manager. On this page, the user gets the main information that is important. It is considered the home page.

### Acceptance Criteria ()
Given that the user is a system administrator or inventory manager, 

1. The user logs into ALTA.
2. The user is navigated to the _Dashboard_ page which contains an overview of audit data over the past month.

### Related Mockups

- Inventory manager logs into ALTA
![image](https://user-images.githubusercontent.com/19229712/102947667-d89c4a00-4491-11eb-9a3a-f133c2af7e22.png)

- Inventory manager arrives on the dashboard upon successful login
<img width="820" alt="dashboard" src="https://user-images.githubusercontent.com/35980232/110572307-25a1cb00-8127-11eb-9c16-fadedd651424.png">

PR #270, opened 2021-03-17
Issue #195 - User Story 5.7

Neeham:
Issue #195

This PR introduces the ability to view the Dashboard in order to have a quick summary of the last 30 days audit. The table shows the Audits and relevant information related to it. The area chart below the table shows the Audit accuracy over the last 30 days.

### DIFF
--- a/client/src/app/components/dashboard/dashboard.component.scss
+++ b/client/src/app/components/dashboard/dashboard.component.scss
@@ -1,9 +1,30 @@
 @import "./src/theme.scss";
 
+:host ::ng-deep .mat-sort-header-container {
+  display: flex;
+  justify-content: center;
+}
+
 table {
+  text-align: center !important;
   width: 100%;
 }
 
 #chart {
-  max-width: 60%;
+  max-width: 100%;
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+.Active {
+  color: red;
+}
+
+.Complete {
+  color: green;
+}
+
+.Pending {
+  color: grey;
 }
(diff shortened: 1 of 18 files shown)


client/src/app/components/dashboard/dashboard.component.scss
```diff
@@ -0,0 +1,47 @@
+@import "./src/theme.scss";
+
+:host ::ng-deep .mat-sort-header-container {
+  display: flex;
+  justify-content: center;
+}
```
NicholasNagy, line 6: Do we absolutely need to use ng-deep here?

NicholasNagy (commented):
I know you're still working on this PR, but please check out my comment :)

client/src/app/components/dashboard/dashboard.component.scss
Neeham: This whole block has been removed as it is no longer needed since width of 100% is being used for the area graph.

Neeham: If we don't use ng-deep it won't apply the CSS this is because ng-deep forces the CSS to be applied to it's child components. Another reason is some components can't access the elements directly because they are outside of this components so ng-deep takes care of that.

NicholasNagy: yes, but previously, this had impacts on other components in the DOM tree. Can you guarantee that nothing will be affected by this change?

Neeham: I see your point and I completely agree. This is why I cannot guarantee that it will not affect other components so in order to avoid this and to guarantee, I will add a parent class to this CSS so that only this specific CSS block is affected by it.

NicholasNagy (approved):
Everything looks good now, thank you for the changes!

PR #283, opened 2021-03-31
Merge audits and templates

Averynder:
This PR merges the functionality brought from the items being filterable to creating the audit with the desired search targets based on an audit template. To start an audit based on one, simply click in the middle of one of the options (boxes) and proceed as if the items for the audit were already selected


makchamp (approved):
Thanks everything works great. 
I just changed the cursor type to be a pointer when the user hovers over the audit template box and updated the migration files.

PR #285, opened 2021-04-04
Extract language out of components to easily change languages

NicholasNagy:
The goal of this PR was to extract out the language out of the HTML in order to have a central place for the language of the system to be easily modified. For each component, a new file was extracted associated to it that handles the functionality and design of the component to handle multiple implementations of different languages.

These components were also subsequently refactored in the last 2 commits to have a central place to modify the language being used instead of handling that on a component basis, as it makes more sense to have a single place to switch languages instead of at a component level, since that would be redundant.


client/src/app/components/review-audit/review-audit.language.ts
```diff
@@ -0,0 +1,52 @@
...
+  title: string;
+  actionButtons: ReviewAuditActionButtons;
+  tableHeaders: ReviewAuditTableHeaders;
+  constructor() {
+    this.title = 'Review Audit';
+    this.actionButtons = {
+      expand: 'Expand All', collapse: 'Collapse All', back: 'Go Back', discard: 'Discard', confirm: 'Confirm', cancel: 'Cancel'
+    };
+    this.tableHeaders = {
+      stock_keeper: 'Stock Keeper', bins: 'Bins', num_parts: 'Number of Parts', init_by: 'Initiated by', init_on: 'Initiated on'
```
Amanda-Kolopa, line 36: Can you please change:

Initiated by -> Initiated By
Initiated on -> Initiated On

Amanda-Kolopa (changes requested):
Awesome work!
Aside from minor capitalization error (added a comment to specify), all looks good to merge :smile:

Amanda-Kolopa (approved):
Thank you for the quick changes!
All looks good to merge :smile:

Issue #67, opened 2020-10-18
User Story 5.6


NicholasNagy:
### User Story: ALTA-5.6

As an inventory manager, I want to be able to comment on an audit report so that I can provide additional information for other inventory managers to review. 

### Description

Inventory managers and stock-keepers should be able to communicate regarding specific audit information. This feature will be useful when informing co-workers of inventory-related topics such as damaged or newly discovered items.

### Acceptance Criteria (AT-5.6 - #279)
Given that the user is logged in as an inventory manager,

1. The user should be redirected to the _Dashboard_ page.
2. The user selects the _Audits_ side menu option.
3. The user is presented with a table containing the list of ongoing and archived audits.
4. The user selects a desired audit and is presented with another table detailing the information of the audit.
5. The user selects the _Comment_ button. The right-side menu appears and displays the conversation attached to the audit.

### Related Mockups 

- Login page
![image](https://user-images.githubusercontent.com/19229712/102693845-82a67880-41eb-11eb-80c3-c881ac759b4d.png)

- Dashboard page
![dashboard](https://user-images.githubusercontent.com/35980232/112053199-6f78b100-8b2a-11eb-8342-c3160b4b099b.png)

- Audit page - displays the list of previous and ongoing audits
![image](https://user-images.githubusercontent.com/19229712/111889898-12122200-89bb-11eb-8392-ef442fb735b6.png)

- Audit Report page - appears once the user selects an audit from the list
![image](https://user-images.githubusercontent.com/19229712/111889933-408ffd00-89bb-11eb-911d-b92d7eecbbb2.png)

- Comment section
![image](https://user-images.githubusercontent.com/19229712/111891575-90c28b80-89ca-11eb-99b0-89fea00c6bfd.png)

Issue #64, opened 2020-10-18
User Story 5.3


NicholasNagy:
### User Story: ALTA-5.3

As an inventory manager, I want to be able to review archived audits so that I can analyze the inventory’s accuracy levels for trends.

### Description

An inventory manager should be able to view a list of archived and ongoing audits to be able to analyze a brief overview or detailed audit results. 

### Acceptance Criteria (AT-5.3 - #277)
Given that the user is logged in as an inventory manager,

1. The user should be redirected to the _Dashboard_ page.
2. The user selects the _Audits_ side menu option.
3. The user is presented with a table containing the list of ongoing and archived audits.
4. The user selects the arrow at the end of a recorded audit. A brief drop-down table lists an overview of the select audit's content.
5. The user selects a desired audit and is presented with another table detailing the information of the audit.

### Related Mockups 
- Login page
![image](https://user-images.githubusercontent.com/19229712/102693845-82a67880-41eb-11eb-80c3-c881ac759b4d.png)

- Dashboard page
![dashboard](https://user-images.githubusercontent.com/35980232/112053199-6f78b100-8b2a-11eb-8342-c3160b4b099b.png)

- Audits page
![Capture d’écran 2021-02-13 à 21 48 39](https://user-images.githubusercontent.com/35980232/107867054-4a1ac800-6e45-11eb-94a6-6bc3c080d34e.png)

- Brief audit overview drop-down option
![image](https://user-images.githubusercontent.com/19229712/112736362-82272780-8f28-11eb-870c-b1b1f40e5a0d.png)

- Audit report page
![Capture d’écran 2021-02-13 à 21 48 47](https://user-images.githubusercontent.com/35980232/107867063-50a93f80-6e45-11eb-81a0-521b48963120.png)

Issue #63, opened 2020-10-18
User Story 5.2


NicholasNagy:
### User Story: ALTA-5.2

As an inventory manager, I want to be able to export the inventory data of a performed audit so that I can use it in the company’s ERP system.

### Description
An inventory manager should be able to generate a CSV file capturing the results of a selected audit to incorporate inventory modifications into their organization's ERP system. 

### Acceptance Criteria (AT-5.2 - #276)
Given that the user is logged in as an inventory manager,

1. The user should be redirected to the _Dashboard_ page.
2. The user selects the _Audits_ side menu option.
3. The user is presented with a table containing the list of ongoing and archived audits.
4. The user selects a desired audit and is presented with another table detailing the information of the audit.
5. The user selects the _Export Audit_ button. The CSV file downloads onto the user's device.

### Related Mockups 

- Login page
![image](https://user-images.githubusercontent.com/19229712/102693845-82a67880-41eb-11eb-80c3-c881ac759b4d.png)

- Dashboard page
![dashboard](https://user-images.githubusercontent.com/35980232/112053199-6f78b100-8b2a-11eb-8342-c3160b4b099b.png)

- Audits page
![Capture d’écran 2021-02-13 à 21 48 39](https://user-images.githubusercontent.com/35980232/107867054-4a1ac800-6e45-11eb-94a6-6bc3c080d34e.png)

- Audit report page
![Capture d’écran 2021-02-13 à 21 48 47](https://user-images.githubusercontent.com/35980232/107867063-50a93f80-6e45-11eb-81a0-521b48963120.png)

- Export audit report
![image](https://user-images.githubusercontent.com/19229712/111891741-632a1200-89cb-11eb-83cf-a610d77e8469.png)

Issue #62, opened 2020-10-18
User Story 5.1


NicholasNagy:
### User Story: ALTA-5.1

As an inventory manager, I want to be able to view an audit report so that I can analyze the discovered discrepancies of a completed audit in a graphical format.

### Acceptance Criteria (AT-5.1 - #275):
Given that the user is logged in as an inventory manager,

1. The user should be redirected to the _Dashboard_ page.
2. The user selects the _Audits_ side menu option.
3. The user is presented with a table containing the list of ongoing and archived audits.
4. The user selects the _Graphs and Trends_ tab on top of the audits table.
5. The user is presented with an interactive graph to view the overall accuracy trends over a specified period of time. 

### Related Mockups 
Login Page 
![image](https://user-images.githubusercontent.com/19229712/102693845-82a67880-41eb-11eb-80c3-c881ac759b4d.png)

Dashboard page 
![dashboard](https://user-images.githubusercontent.com/35980232/112053199-6f78b100-8b2a-11eb-8342-c3160b4b099b.png)

Audit page - displays the list of previous and ongoing audits
![image](https://user-images.githubusercontent.com/19229712/111889898-12122200-89bb-11eb-8392-ef442fb735b6.png)

Graphs and Trends page
<img width="797" alt="Capture d’écran 2021-04-01 à 17 30 12" src="https://user-images.githubusercontent.com/35980232/113356258-f8a09c80-930f-11eb-8764-cb341b19c9f7.png">

PR #287, opened 2021-04-05
Audit report

domludera:
User Stories:  5.1 (#62), 5.2 (#63), 5.3 (#64),  5.6 (#67)


NicholasNagy (commented):
Alright, so under lots of pressures, I realise that a lot of our comments and discussions related to this PR happened offline, so in order to document that, I will include our discussions below:

Firstly, we had issues  with the backend API endpoint for the comments (for US 5.6 #67 ), since we were trying to add a primary key that was unnecessary for comments (as they will never need to be referenced by their primary key), and thus were wasting space. 

The UI of the drill down chart and the UI of the audit report (for US 5.3 #64  ) were also iterated over a bit in order to only show pertinent information useful to understanding the outcome of the audit. Here is the final product, validated by Ajay himself:

image

image

Additionally, we moved the code from a separate branch that needed the code from this one to work effectively (for US 5.2 #63 ) so that we could get immediate feedback from Ajay (for which he mentioned that we will need to change how the report is generated to include the output of actual items found after the audit, not the items from the extract that were used for the audit).

Additionally, changes were asked for the comments, since they originally didn't include the names for each comment box, nor the date and commenting a lot meant having to scroll down to see the newest messages, which wasn't user friendly.

server/login/views.py
```diff
@@ -127,6 +128,10 @@ def save_new_pin(email, user):
     request.data = {'password': first_part + second_part + third_part}
     # For e2e purposes, uncomment this line:
     # request.data = {'password': 'password'}
+    print('\n\n--------------------------')
+    print(request.data)
+    print('--------------------------\n\n')
```
Amanda-Kolopa, line 133: Are these required?
If not, please remove it

NicholasNagy: no, we can remove that

NicholasNagy (approved):
Okay, I looked over the code, including @Amanda-Kolopa 's and I don't see anything new that needs changing. Everything was updated based on our feedback from Ajay, and the user stories are fully completed.

The tests for these changes will be included in a future PR.

For now, it should be good to go. Can you look through my changes @Amanda-Kolopa , so that we can validate each other's code?

Amanda-Kolopa (approved):
Thank you @NicholasNagy for reviewing the other aspects of the code!
All is working well on my end :smile:

PR #291, opened 2021-04-10
Comment component refactoring

domludera:
This refactoring extracts the comment component from the audit report to be its own separate entity. We have extracted the class, methods, and variables. Doing so will enable the comment component to be independent of the audit report page. This helps decouple and both features and favors reuse. 

It allows moving members of an existing class to a new class. It is helpful by replacing a single class that is responsible for multiple tasks with several classes each having a single responsibility.


NicholasNagy (approved):
Thank you dom, since we had discussed the design for this together, with the team, I don't have any design feedback, nor do I see any bugs in this.

We will merge it into the other branch for further review!

